PR #121, opened 2024-03-14
wip: feat: add an option to use whitelist for the students

zhmurov:
## Description

This add an option to connect a spreadsheet with pre-defined list of students that are allowed to register in GitLab. If this option is selected, the username is checked upon signup. If the username is whitelisted, the user is allowed, otherwise corresponding error message is printed. Note that if user is already registered, they can login ignoring the whitelist.

To select using the whitelist, one needs to add following options to .env file:

USE_WHITELIST=True
GDOC_WHITELIST_ID=<spreadshhet_ID>
GDOC_WHITELIST_SHEET=0

## Motivation and Context

Currently, anyone knowing the secret word can register. Since the 'secret' can be easily passed from one person to the other, another mechanism of restricting the access is needed.

Also, when people are asked to fill in their full name and email while registering in manytask, they can use different transliteration, type their nickname instead of full name, ignore typing last name, use different email address, etc. This makes it hard to match peoples across different databases, e.g. when transfering grades from Manytask to LMS. With this patch, the full name and email are taken from the whitelist, which ensures that we know who is who.

Related to 
* #
* #
* 

Closes #

## Contribution Checklist

- [ ] I have read and agreed on the [CONTRIBUTING](../CONTRIBUTING.md) and [CODE OF CONDUCT](../CODE_OF_CONDUCT.md) documents.
- [ ] I have run linters, type checks, import sorting and tests locally.
- [ ] I have added comments to code in hard-to-understand areas.
- [ ] I have added 100% coverage tests for the new code (or aren't needed).
- [ ] I have added docs for the new code (or aren't needed).
- [ ] I have used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) title: `tag(scope): description`.

### DIFF
--- a/manytask/course.py
+++ b/manytask/course.py
@@ -104,5 +104,9 @@ class Course:
         return self.googledoc_api.fetch_rating_table()
     
     @property
+    def use_whitelist(self) -> bool:
+        return self.googledoc_api.use_whitelist
+    
+    @property
     def whitelist_table(self) -> "gdoc.WhitelistTable":
         return self.googledoc_api.fetch_whitelist_table()
--- a/manytask/course.py
+++ b/manytask/course.py
@@ -102,3 +102,7 @@ class Course:
     @property
     def rating_table(self) -> "gdoc.RatingTable":
         return self.googledoc_api.fetch_rating_table()
+    
+    @property
+    def whitelist_table(self) -> "gdoc.WhitelistTable":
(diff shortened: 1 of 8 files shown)


akostrikov:
Lets get to new year withoud hanging reviews.
On the next meeting we should decide how to rewrite the code and do it idiomatic.

PR #158, opened 2024-08-15
fix: mypy errors

MoskalenkoViktor:
After dependencies were updated, some of the function calls were marked as type-unsafe by mypy. This change fixes the problem by explicitly assigning types to the variables where needed.


zhmurov:
It does have the 'fix' label. I think these checks are very discouraging and must be disabled...

zhmurov (approved):
LGTM

PR #168, opened 2024-09-10
fix: Allow non-integer number format for score reporting

zhmurov:
Current version of a checker returns score as a string that represents float number, which is casted directly to int. This cause a value error.

The proposed fis allow to push integers or floats in the score string, with the following behavior:
1. If integer is passed, it is taken as final score for the task
2. If a float less than zero is passed, set the score to 0
3. If a float larger than 2.0 is passed, report and error
4. If a float between 0.0 and 2.0 is passed, use it a fractional score, i.e. multiply it by the maximum score for the task.

### DIFF
--- a/manytask/api.py
+++ b/manytask/api.py
@@ -131,14 +131,6 @@ def report_score() -> ResponseReturnValue:
     if "check_deadline" in request.form:
         check_deadline = request.form["check_deadline"] is True or request.form["check_deadline"] == "True"
 
-    reported_score: int | None = None
-    if "score" in request.form:
-        score_str = request.form["score"]
-        try:
-            reported_score = int(score_str)
-        except ValueError:
-            return f"Cannot parse `score` <{reported_score}> to int`", 400
-
     submit_time = None
     submit_time_str = None
     if "submit_time" in request.form:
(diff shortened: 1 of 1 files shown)


manytask/api.py
```diff
@@ -160,6 +152,21 @@ def report_score() -> ResponseReturnValue:
...
+    reported_score: int | None = None
+    if "score" in request.form:
+        score_str = request.form["score"]
+        try:
+            if score_str.isdigit():
+                reported_score = int(score_str)
+            elif float(score_str) < 0.0:
+                reported_score = 0
+            elif float(score_str) > 1.0:
+                reported_score = int(round(float(score_str)))
```
k4black, line 164: Bonus scores can apply, lets' do the following 
* if int = tackle as final score
* if not = tackle as float * max_score
* if float check <= 2.0 and error if too big

k4black (changes requested):

manytask/api.py
```diff
@@ -160,6 +152,21 @@ def report_score() -> ResponseReturnValue:
...
+        score_str = request.form["score"]
+        try:
+            if score_str.isdigit():
+                reported_score = int(score_str)
+            elif float(score_str) < 0.0:
+                reported_score = 0
+            elif float(score_str) > 2.0:
+                return f"Reported `score` <{reported_score}> is too large. Should be ineger or float between 0.0 and 2.0`", 400
+            else:
+                reported_score = int(round(float(score_str) * task.score))
```
k4black, line 166: FYI round return int
Screenshot 2024-09-10 at 10 00 43

k4black:
LGTM

manytask/api.py
zhmurov: Fixed

PR #190, opened 2024-11-24
refactor: separate and make abstract gdoc api

gagarinkomar:
### Separate GoogleDocApi into viewer and storage api

Separate GoogleDocApi into GoogleDocViewerApi and GoogleDocStorageApi.
Separate RatingTable into ViewerRatingTable and StorageRatingTable.
In Course instead of googledoc_api using viewer_api and storage_api.

### Make separated classes abstract

Create abstract classes(ViewStoreApi, RatingApiBase, ViewerRatingApi, StorageRatingApi).
ViewStoreApi using in initial classes like GoogleDocViewerApi/GoogleDocStorageApi for having fetch_rating_api method.
RatingApiBase using in viewer/storage classes like ViewerRatingTable/StorageRatingTable for having methods which work with cache.
ViewerRatingApi/StorageRatingApi also unsing in viewer/storage classes for having view/store data methods.

### DIFF
--- a/manytask/abstract.py
+++ b/manytask/abstract.py
@@ -1,0 +1,59 @@
+from abc import ABC, abstractmethod
+from typing import Any, Callable
+
+from config import ManytaskDeadlinesConfig
+from glab import Student
+
+
+class ViewerApi(ABC):
+    @abstractmethod
+    def get_spreadsheet_url(self) -> str:
+        ...
+
+
+class StorageApi(ABC):
+    @abstractmethod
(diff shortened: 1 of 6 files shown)


feldsherov (changes requested):
LGTM overall. 

1. Please take a look onto large comment about GDocInit.
2. Now your PR has 3 commits. Can we squash them to contain only this final state.
3. We need to check that this is working before merge. Do you have a dev instance to check that we are still ok?

PR #193, opened 2024-12-02
feat: add db models for DatabaseApi

gagarinkomar:
Create flask sqlalchemy orm models in models.py.
Add simple tests for models in tests/test_models.py.
Add necessary modules to requirements.txt.

[Database scheme](https://drawsql.app/teams/gagarinkomar/diagrams/copy-of-manytask-2)

### DIFF
--- a/manytask/models.py
+++ b/manytask/models.py
@@ -1,0 +1,123 @@
+from sqlalchemy.orm import Mapped, mapped_column, DeclarativeBase, relationship
+from sqlalchemy import ForeignKey, JSON, UniqueConstraint
+from typing import Optional, List
+from datetime import datetime, timezone
+
+
+class Base(DeclarativeBase):
+    pass
+
+
+class User(Base):
+    __tablename__ = 'users'
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    username: Mapped[str]
+    gitlab_instance_host: Mapped[str] = mapped_column(default='gitlab.manytask.org')
+    is_manytask_admin: Mapped[bool] = mapped_column(default=False)
+
+    __table_args__ = (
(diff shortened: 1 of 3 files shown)


manytask/models.py
```diff
@@ -0,0 +1,123 @@
...
+class Base(DeclarativeBase):
+    pass
+
+
+class User(Base):
+    __tablename__ = 'users'
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    username: Mapped[str]
+    gitlab_instance_host: Mapped[str] = mapped_column(default='gitlab.manytask.org')
```
feldsherov, line 16: Are you sure that we want any default here? Having a default of domain on the model level feels wierd. But I could be wrong.

If you like it, let's talk.
If you are not attached to it, let's drop it from here.

manytask/models.py
```diff
@@ -0,0 +1,123 @@
...
+    is_manytask_admin: Mapped[bool] = mapped_column(default=False)
+
+    __table_args__ = (
+        UniqueConstraint('username', 'gitlab_instance_host', name='_username_gitlab_instance_uc'),
```
feldsherov, line 20: If formatting right here? Line is too long.

manytask/models.py
```diff
@@ -0,0 +1,123 @@
...
+    __table_args__ = (
+        UniqueConstraint('user_id', 'task_id', name='_user_task_uc'),
+    )
+
+    # relationships
+    user: Mapped['User'] = relationship(back_populates='grades')
+    task: Mapped['Task'] = relationship(back_populates='grades')
+
+
+# Is it needed?
```
feldsherov, line 114: Nope. Let's drop it.

feldsherov (changes requested):
LGTM. I left a few technical comments. Please fix them and let's merge.

feldsherov (approved):

PR #199, opened 2024-12-13
feat: create database api class

gagarinkomar:
Created DataBaseApi class derived from abstract StorageApi.
Add classmethods in models for interaction with database models from api class.
Also changed a few models and tests for them.
Add database parameters for .env.example. In main use DataBaseApi instead of GoogleDocApi if USE_DATABASE_AS_STORAGE=true.

Need to test this more and better.

### DIFF
--- a/manytask/database.py
+++ b/manytask/database.py
@@ -1,0 +1,155 @@
+from __future__ import annotations
+
+import json
+import logging
+from datetime import datetime, timezone
+from typing import Any, Callable
+
+from sqlalchemy import create_engine
+from sqlalchemy.exc import NoResultFound
+from sqlalchemy.orm import Session
+
+from . import models
+from .abstract import StorageApi
+from .config import ManytaskDeadlinesConfig
+from .glab import Student
+
+
+logger = logging.getLogger(__name__)
+
+
+class DataBaseApi(StorageApi):
+    def __init__(
+        self,
+        database_url: str,
+        course_name: str,
+        gitlab_instance_host: str,
+        registration_secret: str,
+        show_allscores: bool
+    ):
+        """
+        :param database_url:
+        :param course_name:
+        :param gitlab_instance_host:
+        :param registration_secret:
+        :param show_allscores:
+        """
+
+        self.engine = create_engine(database_url, echo=False)
+
+        with Session(self.engine) as session:
+            self.course_name = models.Course.update_or_create(
+                session, course_name, gitlab_instance_host, registration_secret, show_allscores).name
+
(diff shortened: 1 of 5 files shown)


manytask/database.py
```diff
@@ -0,0 +1,155 @@
+from __future__ import annotations
```
feldsherov, line 1: Which version of Python do we run?
Are annotations just typehints? I assume we have them without future import.

manytask/models.py
feldsherov, line 1: Please move schema update to separate commit.

manytask/models.py
```diff
@@ -15,7 +17,7 @@ class User(Base):
     id: Mapped[int] = mapped_column(primary_key=True)
     username: Mapped[str]
     gitlab_instance_host: Mapped[str]
-    is_manytask_admin: Mapped[bool] = mapped_column(default=False)
+    # is_manytask_admin: Mapped[bool] = mapped_column(default=False) TODO: now it unused in model
```
feldsherov, line 20: IMO this is ok if it is unused and still here. But keeping commented code is antipattern.

Please uncomment or drop from the code.

manytask/models.py
```diff
@@ -26,32 +28,132 @@ class User(Base):
     )
 
     # relationships
-    users_on_courses: Mapped[List['UserOnCourse']] = relationship(
+    users_on_courses: DynamicMapped[List['UserOnCourse']] = relationship(
         back_populates='user', lazy='dynamic')
-    grades: Mapped[List['Grade']] = relationship(back_populates='user', lazy='dynamic')
+
+    @classmethod
+    def get_or_create(
```
feldsherov, line 35: get_or_create/get_by_name etc class methods are business logic, they are not belong to model.
Model is description of how database maps to Python object.
It should be free from logic about how to query anything from DB.

I see two options:

1. Create a bunch of helpers [like this](https://gist.github.com/jleclanche/f53ab91184632d717b228ec782d36dd8) and move all logic to DataBaseApi.
2. Create a few wrapper classes which will contain model objects as filed and implement all business logic you need.

I would prefer 1.

manytask/main.py
feldsherov, line 1: Main changes is again separate commit.

manytask/database.py
feldsherov, line 1: Please cover this file with tests. I assume we can achieve mostly full coverage for this class.

feldsherov (changes requested):

feldsherov:
@gagarinkomar should we close this PR in favor of other three?

PR #200, opened 2024-12-16
refactor: edit models

gagarinkomar:
Changed lazy='dynamic' to DynamicMapped, it is the same thing but more convenient.
In Grade instead of User relationship using UserOnCourse.
Swapped TaskGroup and Task for convenience(link to course now in TaskGroup instead of Task).
Removed unused fields: is_manytask_admin, is_course_admin.
Add gitlab_instance_host field to course and validation gitlab_instance_host equal when creating UserOnCourse.
Also edited test_models.py

Current scheme:
https://drawsql.app/teams/gagarinkomar/diagrams/copy-of-copy-of-manytask

### DIFF
--- a/manytask/models.py
+++ b/manytask/models.py
@@ -1,9 +1,13 @@
+import logging
 from datetime import datetime
 from typing import Optional
 
 from sqlalchemy import JSON, ForeignKey, UniqueConstraint, event, func
 from sqlalchemy.engine import Connection
 from sqlalchemy.orm import DeclarativeBase, DynamicMapped, Mapped, Mapper, Session, mapped_column, relationship
+
+
+logger = logging.getLogger(__name__)
 
 
 class Base(DeclarativeBase):
@@ -88,7 +92,9 @@ def validate_gitlab_instance_host(mapper: Mapper[UserOnCourse], connection: Conn
             course = target.course
         else:
             course = session.query(Course).filter_by(id=target.course_id).one()
-    except Exception:
+    except Exception as e:  # TODO: fix swallowing exception
+        logger.warning(
+            f"Swallowing exception in 'before_insert' event in UserOnCourse model: {repr(e)}")
         return
 
     if user.gitlab_instance_host != course.gitlab_instance_host:
--- a/manytask/models.py
+++ b/manytask/models.py
@@ -85,12 +85,12 @@ def validate_gitlab_instance_host(mapper: Mapper[UserOnCourse], connection: Conn
     try:
         if target.user:
             user = target.user
-        else:
+        else:  # target missing user if we gave user_id or nothing
             user = session.query(User).filter_by(id=target.user_id).one()
 
         if target.course:
             course = target.course
-        else:
+        else:  # target missing course if we gave course_id or nothing
             course = session.query(Course).filter_by(id=target.course_id).one()
     except Exception as e:  # TODO: fix swallowing exception
         logger.warning(
--- a/manytask/models.py
+++ b/manytask/models.py
@@ -1,7 +1,7 @@
-from datetime import datetime, timezone
+from datetime import datetime
 from typing import Optional
 
-from sqlalchemy import JSON, ForeignKey, UniqueConstraint, event
+from sqlalchemy import JSON, ForeignKey, UniqueConstraint, event, func
 from sqlalchemy.engine import Connection
 from sqlalchemy.orm import DeclarativeBase, DynamicMapped, Mapped, Mapper, Session, mapped_column, relationship
 
@@ -27,7 +27,9 @@ class User(Base):
 
     # relationships
     users_on_courses: DynamicMapped['UserOnCourse'] = relationship(
-        back_populates='user')
+        back_populates='user',
+        cascade='all, delete-orphan'
+    )
 
 
 class Course(Base):
@@ -41,9 +43,13 @@ class Course(Base):
 
     # relationships
     task_groups: DynamicMapped['TaskGroup'] = relationship(
-        back_populates='course')
+        back_populates='course',
+        cascade='all, delete-orphan'
+    )
     users_on_courses: DynamicMapped['UserOnCourse'] = relationship(
-        back_populates='course')
+        back_populates='course',
+        cascade='all, delete-orphan'
+    )
 
 
 class UserOnCourse(Base):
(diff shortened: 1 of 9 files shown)


manytask/models.py
```diff
@@ -62,13 +62,36 @@ class UserOnCourse(Base):
...
+            user = target.user
+        else:
+            user = session.query(User).filter_by(id=target.user_id).one()
+
+        if target.course:
+            course = target.course
+        else:
+            course = session.query(Course).filter_by(id=target.course_id).one()
+    except Exception:
+        return
```
feldsherov, line 84: Why do you need to swallow exception here?

You are catching exception and after that just return, this is antipattern, do we have a good reason for doing this?

manytask/models.py
```diff
@@ -62,13 +62,36 @@ class UserOnCourse(Base):
...
+
+@event.listens_for(UserOnCourse, 'before_insert')
+def validate_gitlab_instance_host(mapper: Mapper[UserOnCourse], connection: Connection, target: UserOnCourse) -> None:
+    session = Session(bind=connection)
+
+    try:
+        if target.user:
+            user = target.user
+        else:
+            user = session.query(User).filter_by(id=target.user_id).one()
```
feldsherov, line 77: When do we expect to have missing user in target?
Drop a comment, why can it happen?

manytask/models.py
```diff
@@ -62,13 +62,36 @@ class UserOnCourse(Base):
...
+
+        if target.course:
```
feldsherov, line 79: Same question about missing course.

feldsherov (changes requested):

manytask/models.py
```diff
@@ -26,23 +26,24 @@ class User(Base):
...
+    gitlab_instance_host: Mapped[str]
     registration_secret: Mapped[str]
     show_allscores: Mapped[bool] = mapped_column(default=False)
 
     # relationships
-    tasks: Mapped[List['Task']] = relationship(back_populates='course', lazy='dynamic')
-    users_on_courses: Mapped[List['UserOnCourse']] = relationship(
-        back_populates='course', lazy='dynamic')
+    task_groups: DynamicMapped['TaskGroup'] = relationship(
+        back_populates='course')
```
prawwtocol, line 44: don't we need to set up cascade options to properly handle orphaned records?
something like
`cascade="all, delete-orphan"`

manytask/models.py
```diff
@@ -26,23 +26,24 @@ class User(Base):
...
+    users_on_courses: DynamicMapped['UserOnCourse'] = relationship(
+        back_populates='course')
```
prawwtocol, line 46: same question here

manytask/models.py
```diff
@@ -79,40 +102,41 @@ class TaskGroup(Base):
...
     __tablename__ = 'grades'
 
     id: Mapped[int] = mapped_column(primary_key=True)
-    user_id: Mapped[int] = mapped_column(ForeignKey(User.id))
+    user_on_course_id: Mapped[int] = mapped_column(ForeignKey(UserOnCourse.id))
     task_id: Mapped[int] = mapped_column(ForeignKey(Task.id))
-    score: Mapped[int]
-    submit_date: Mapped[datetime]
+    score: Mapped[int] = mapped_column(default=0)
+    last_submit_date: Mapped[datetime] = mapped_column(default=lambda: datetime.now(timezone.utc))
```
prawwtocol, line 134: why not `func.now()`?

tests/test_models.py
```diff
@@ -154,117 +172,154 @@ def test_task_group(session):
...
+    session.add_all([user, course, user_on_course, task_group, task])
     session.commit()
 
     grade = Grade(
-        user=user,
+        user_on_course=user_on_course,
         task=task,
         score=77,
-        submit_date=datetime.now(timezone.utc)
+        last_submit_date=datetime.now(timezone.utc)
```
prawwtocol, line 208: to ensure consistency, create a fixture, for e.g. `current_time` and specify a fixed datetime there

tests/test_models.py
```diff
@@ -133,9 +143,14 @@ def test_user_on_course_unique_ids(session):
 
 
 def test_deadline(session):
+    course = Course(name="course0001", registration_secret="secret",
+                    gitlab_instance_host='gitlab.inst.org')
+    session.add(course)
+    session.commit()
+
     deadline = Deadline(data={"test_key": "test_value"})
```
prawwtocol, line 151: I think we should also test invalid json and empty data

manytask/models.py
gagarinkomar: Because if there is a problem with finding user or course, we don't want to think about exceptions here. We swallow the exception so that sqlalchemy.orm raises the necessary exception for us when trying to create an object itself.

manytask/models.py
gagarinkomar:
```
user_on_course = UserOnCourse(
    user_id=1,
    course=course,
    repo_name="user_repo"
)
```

If we create an object like this, we have user_id instead of user

manytask/models.py
gagarinkomar: Similar example

manytask/models.py
gagarinkomar: Added cascade options and tests for them

manytask/models.py
gagarinkomar: Changed this

tests/test_models.py
gagarinkomar: Created fixture and refactor code parts with current time usage

tests/test_models.py
gagarinkomar: Add test_deadline_date for test it

manytask/models.py
prawwtocol: @gagarinkomar yeah I think the point was that you were advised to include that exact comment in code -- so that anyone reading this method would understand why target.user could be None

manytask/models.py
prawwtocol: @gagarinkomar I think the worst part of this is catching a The Most General Exception Posssible (TM). Maybe NoResultFound would be better?

akostrikov (approved):
LGTM

manytask/models.py
akostrikov: @gagarinkomar can you add logging here and a todo?
Seems like a funny place to swallow errors later.

akostrikov: https://github.com/manytask/manytask/issues/219 added issue

PR #201, opened 2024-12-16
feat: create database api class

gagarinkomar:
Created DataBaseApi class derived from abstract StorageApi.
Created get_or_create, update_or_create, etc. methods like django funcs.
Add tests to this class(tests/test_db_api.py).

### DIFF
--- a/manytask/database.py
+++ b/manytask/database.py
@@ -1,8 +1,8 @@
 import logging
 from datetime import datetime, timedelta, timezone
-from typing import Any, Callable, Iterable, List, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Iterable, List, Optional, Type, TypeVar
 
-from sqlalchemy import create_engine
+from sqlalchemy import and_, create_engine
 from sqlalchemy.exc import IntegrityError, NoResultFound
 from sqlalchemy.orm import Session
 from sqlalchemy.sql.functions import func
@@ -17,7 +17,9 @@ logger = logging.getLogger(__name__)
 
 
 class DataBaseApi(StorageApi):
-    T = TypeVar('T', bound=models.Base)
+    """Class for interacting with a database with the StorageApi functionality"""
+
+    ModelType = TypeVar('ModelType', bound=models.Base)
 
     def __init__(
         self,
@@ -27,25 +29,28 @@ class DataBaseApi(StorageApi):
         registration_secret: str,
         show_allscores: bool
     ):
-        """
-        :param database_url:
-        :param course_name:
-        :param gitlab_instance_host:
-        :param registration_secret:
-        :param show_allscores:
+        """Constructor of DataBaseApi class
+
+        :param database_url: full url for database connection
+        :param course_name: unique course name
+        :param gitlab_instance_host: gitlab instance host url
+        :param registration_secret: secret to registering for course
+        :param show_allscores: flag for showing results to all users
         """
 
         self.engine = create_engine(database_url, echo=False)
 
         with Session(self.engine) as session:
             try:
-                course = self.get(session, models.Course, name=course_name)
+                course = self._get(session, models.Course, name=course_name)
                 if course.gitlab_instance_host != gitlab_instance_host:
-                    AttributeError("Can't update gitlab_instance_host param on created course")
+                    raise AttributeError(
+                        "Can't update gitlab_instance_host param on created course")
             except NoResultFound:
                 pass
 
-            self.course_name = self.update_or_create(
+            self.course_name = course_name
+            self._update_or_create(
                 session,
                 models.Course,
                 defaults={
@@ -53,33 +58,25 @@ class DataBaseApi(StorageApi):
                     'registration_secret': registration_secret,
                     'show_allscores': show_allscores
                 },
-                name=course_name
-            ).name
+                name=self.course_name
+            )
 
     def get_scores(
         self,
         username: str,
     ) -> dict[str, int]:
+        """Method for getting all user scores
+
+        :param username: student username
+
+        :return: dict with the names of tasks and their scores
+        """
+
         with Session(self.engine) as session:
-            try:
-                course = self.get(session, models.Course, name=self.course_name)
-                user = self.get(
-                    session,
-                    models.User,
-                    username=username,
-                    gitlab_instance_host=course.gitlab_instance_host
-                )
+            grades = self._get_scores(session, username, only_bonus=False)
 
-                user_on_course = self.get(
-                    session,
-                    models.UserOnCourse,
-                    user_id=user.id,
-                    course_id=course.id
-                )
-            except NoResultFound:
+            if grades is None:
                 return {}
-
-            grades = self.get_all_grades(user_on_course, only_bonus=False)
 
             scores: dict[str, int] = {}
             for grade in grades:
@@ -91,32 +88,29 @@ class DataBaseApi(StorageApi):
         self,
         username: str,
     ) -> int:
+        """Method for getting user's total bonus score
+
+        :param username: student username
+
+        :return: user's total bonus score
+        """
+
         with Session(self.engine) as session:
-            try:
-                course = self.get(session, models.Course, name=self.course_name)
-                user = self.get(
-                    session,
-                    models.User,
-                    username=username,
-                    gitlab_instance_host=course.gitlab_instance_host
-                )
+            grades = self._get_scores(session, username, only_bonus=True)
 
-                user_on_course = self.get(
-                    session,
-                    models.UserOnCourse,
-                    user_id=user.id,
-                    course_id=course.id
-                )
-            except NoResultFound:
-                return 0
-
-            grades = self.get_all_grades(user_on_course, only_bonus=True)
+        if grades is None:
+            return 0
 
         return sum([grade.score for grade in grades])
 
     def get_all_scores(self) -> dict[str, dict[str, int]]:
+        """Method for getting all scores for all users
+
+        :return: dict with usernames and all their scores
+        """
+
         with Session(self.engine) as session:
-            all_users = self.get_all_users(session, self.course_name)
+            all_users = self._get_all_users(session, self.course_name)
 
         all_scores: dict[str, dict[str, int]] = {}
         for username in all_users:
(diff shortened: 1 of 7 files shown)


manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+
+            grades = self.get_all_grades(user_on_course, only_bonus=False)
+
+            scores: dict[str, int] = {}
+            for grade in grades:
+                scores[grade.task.name] = grade.score
+
+        return scores
+
+    def get_bonus_score(
```
zhmurov, line 90: This and get_scores look remarkably similar. Suggest private method get_scores() with ad extra boolean argument

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+    def update_cached_scores(self) -> None:
+        return None
+
+    def store_score(
+        self,
+        student: Student,
+        task_name: str,
+        update_fn: Callable[..., Any],
+    ) -> int:
+        flags = ''  # TODO: in GoogleDocApi imported from google table, they used to increase the deadline for the user
```
zhmurov, line 155: Suggest creating an issue

tests/test_db_api.py
```diff
@@ -0,0 +1,400 @@
...
+    Base.metadata.drop_all(engine)
+
+
+@pytest.fixture
+def session(engine, tables):
+    with Session(engine) as session:
+        yield session
+
+
+@pytest.fixture
```
zhmurov, line 37: If it is not much extra work, suggest creating fixtures for google doc api

tests/test_db_api.py
```diff
@@ -0,0 +1,400 @@
...
+
+from manytask.config import ManytaskDeadlinesConfig
+from manytask.database import DataBaseApi
+from manytask.glab import Student
+from manytask.models import Base, Course, Deadline, Grade, Task, TaskGroup, User, UserOnCourse
+
+
+SQLALCHEMY_DATABASE_URL = 'sqlite:///file::memory:?cache=shared'
+DEADLINES_CONFIG_FILE_1 = 'tests/.deadlines.test.yml'  # part of manytask config file
+DEADLINES_CONFIG_FILE_2 = 'tests/.deadlines.test2.yml'  # part of manytask config file
```
zhmurov, line 16: This does not look right to me. It seems that having an array of config files is a better solution

tests/test_db_api.py
```diff
@@ -0,0 +1,400 @@
...
+        database_url=SQLALCHEMY_DATABASE_URL,
+        course_name="Test Course",
+        gitlab_instance_host="gitlab.test.com",
+        registration_secret="secret",
+        show_allscores=True
+    )
+
+
+@pytest.fixture
+def db_api2(tables):
```
zhmurov, line 49: This suggest adding db_api3, db_api_new, etc if we decide to make more test-cases.

zhmurov (commented):
Nice work! I think my suggestions on improving testing infrastructure can be addressed later (but we need an issue if decide to do so)

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+    def __init__(
+        self,
+        database_url: str,
+        course_name: str,
+        gitlab_instance_host: str,
+        registration_secret: str,
+        show_allscores: bool
+    ):
+        """
+        :param database_url:
```
feldsherov, line 31: Please add comment on what every param mean, or drop the doc string, it is useless now...

I suspect linter forced you to add the docstring, so we have no other way, but adding the proper doc string here.

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+from .abstract import StorageApi
+from .config import ManytaskDeadlinesConfig
+from .glab import Student
+
+
+logger = logging.getLogger(__name__)
+
+
+class DataBaseApi(StorageApi):
+    T = TypeVar('T', bound=models.Base)
```
feldsherov, line 20: Can we get more descriptive name. 

Like ModelType, or similar.

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+
+        with Session(self.engine) as session:
+            try:
+                course = self.get(session, models.Course, name=course_name)
+                if course.gitlab_instance_host != gitlab_instance_host:
+                    AttributeError("Can't update gitlab_instance_host param on created course")
+            except NoResultFound:
+                pass
+
+            self.course_name = self.update_or_create(
```
feldsherov, line 48: What is your intent here?

Create the course or get it's name?

In any way I would do:
```
self.course_name = course_name
self.create_or_update(...)
```

manytask/database.py
feldsherov: +1

manytask/database.py
feldsherov, line 1: Can we add underscore to all non StorageApi methods.

This is a convention of marking private methods like this:
https://stackoverflow.com/questions/15047122/python-private-function-coding-convention

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+                    tasks_stats[task.name] = self.get_task_submits_count(
+                        session, task.id) / users_on_courses_count
+
+        return tasks_stats
+
+    def get_scores_update_timestamp(self) -> str:
+        return str(datetime.now(timezone.utc))
+
+    def update_cached_scores(self) -> None:
+        return None
```
feldsherov, line 147: Just `return` will return None.

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+from . import models
+from .abstract import StorageApi
+from .config import ManytaskDeadlinesConfig
+from .glab import Student
+
+
+logger = logging.getLogger(__name__)
+
+
+class DataBaseApi(StorageApi):
```
feldsherov, line 19: Please add doc string to the class and to all public methods.

tests/test_db_api.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+@pytest.fixture
+def session(engine, tables):
+    with Session(engine) as session:
+        yield session
+
+
+@pytest.fixture
+def db_api(tables):
```
feldsherov, line 38: Just 1 and 2 in names are hardly readable, maybe something like:

`first_course_db_api`
`second_course_db_api`

manytask/database.py
feldsherov, line 1: You are using `dict[str,, int]` and other dicts a lot in this file.
This makes code hard to get, to reduce this you can add good comments to the code.

Better to avoid usage of raw dicts, when possible. I.e. when it is not part of public interface which we already have.

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+                            'is_bonus': task.is_bonus
+                        },
+                        name=task.name,
+                        group_id=task_group.id
+                    )
+
+    @staticmethod
+    def serialize_deadline_data(
+        start: datetime,
+        steps: dict[float, Union[datetime, timedelta]],
```
feldsherov, line 246: Here and in other places. Prefer | syntax to Union.

Like: `steps: dict[float, datetime | timedelta],`

tests/test_db_api.py
```diff
@@ -0,0 +1,400 @@
...
+                     'task_3_0': 0.0, 'task_3_1': 0.0, 'task_3_2': 0.0,
+                     'task_4_0': 0.0, 'task_4_1': 0.0, 'task_4_2': 0.0}
+    assert all_scores == {'user1': {'task_1_0': 99}, 'user3': {'task_1_1': 7}}
+    assert bonus_score1 == 0
+    assert scores1 == {'task_1_0': 99}
+    assert bonus_score3 == 0
+    assert scores3 == {'task_1_1': 7}
+
+
+def test_first_course_again(db_api):
```
feldsherov, line 383: Test shouldn't rely on the order of the tests execution.


Seem like we are doing this now....

tests/test_db_api.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+SQLALCHEMY_DATABASE_URL = 'sqlite:///file::memory:?cache=shared'
+DEADLINES_CONFIG_FILE_1 = 'tests/.deadlines.test.yml'  # part of manytask config file
+DEADLINES_CONFIG_FILE_2 = 'tests/.deadlines.test2.yml'  # part of manytask config file
+
+
+@pytest.fixture(scope='module')
+def engine():
+    return create_engine(SQLALCHEMY_DATABASE_URL, echo=False)
```
feldsherov, line 21: We have one db on all tests, this leads to tests see changes of each other.
This is unmaintanable, by same reason as global variable in the code.

Can we:
1. Have helper function which populate DB with test content.
2. Drop state of db or recreate sqlite for every test.

feldsherov (dismissed):

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+
+        return session.query(func.count(models.UserOnCourse.id)).filter_by(course_id=course.id).one()[0]
+
+    @staticmethod
+    def get_task_submits_count(
+        session: Session,
+        task_id: int,
+    ) -> int:
+        return session.query(func.count(models.Grade.id)).filter(
+            models.Grade.task_id == task_id and models.Grade.score > 0).one()[0]
```
prawwtocol, line 424: My understanding that it's either bad practice or downright incorrect to use python's built-in and operator in sqlalchemy queries. Instead you should use 

`from sqlalchemy import and_`

You can also just pass 2 arguments to `filter()` and they will be implicitly ANDed, but explicit is better

@feldsherov

manytask/database.py
```diff
@@ -0,0 +1,424 @@
...
+                if users_on_courses_count == 0:
+                    tasks_stats[task.name] = 0
+                else:
+                    tasks_stats[task.name] = self.get_task_submits_count(
+                        session, task.id) / users_on_courses_count
+
+        return tasks_stats
+
+    def get_scores_update_timestamp(self) -> str:
+        return str(datetime.now(timezone.utc))
```
prawwtocol, line 144: This output can vary based on python version or system locale. It's vetter to use `datetime.isoformat()`

tests/test_db_api.py
prawwtocol: Strongly +1

akostrikov (approved):
LGTM

manytask/database.py
gagarinkomar: Created [issue](https://github.com/manytask/manytask/issues/220)

PR #221, opened 2024-12-25
feat: add auto creation of tables in database

gagarinkomar:
Added CREATE_TABLES_IF_NOT_EXIST flag in env.
If flag is True and tables don't exist, in DataBaseApi tables will be created.
Added test for it.


akostrikov (approved):
I`ll add flag usage later for extracted method.

PR #222, opened 2024-12-27
chore: enhance Docker configuration for production environment

prawwtocol:
- Added DNS settings to the docker-compose file for improved network resolution.
- Included extra_hosts entries for Yandex.Cloud database instances in the docker-compose file.
- Updated the Dockerfile to set up Yandex.Cloud certificate for secure database connections.

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -65,3 +65,10 @@ services:
         volumes:
             - cache:/cache
             - solutions:/solutions
+        dns:
+            - 8.8.8.8
+            - 8.8.4.4
+        extra_hosts:
+            - "rc1b-ccszhe0ozcriehwo.mdb.yandexcloud.net:10.129.0.12"
+            - "rc1b-gomhenwov90duxxz.mdb.yandexcloud.net:10.129.0.35"
+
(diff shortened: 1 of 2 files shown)


docker-compose.production.yml
```diff
@@ -65,3 +65,10 @@ services:
         volumes:
             - cache:/cache
             - solutions:/solutions
+        dns:
+            - 8.8.8.8
+            - 8.8.4.4
+        extra_hosts:
```
prawwtocol, line 71: this is something very weird, but without this specific mapping, I get

```
manytask          | sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) could not translate host name "some-secret-hostname.mdb.yandexcloud.net" to address: Name does not resolve
manytask          | could not translate host name "some-secret-hostname.mdb.yandexcloud.net" to address: Name does not
```

prawwtocol: maybe that's because we're using an alpine image (`FROM python:3.13-alpine as builder`) which doesnt have some network functions??

akostrikov: Seems that it is related to Yandex Cloud security policies?

PR #242, opened 2025-01-20
feat: enhance development workflow with pre-commit hooks and Makefile

prawwtocol:


prawwtocol:
related to: #241

prawwtocol:
doesnt check `tests/` dir with mypy, because https://github.com/manytask/manytask/blob/main/pyproject.toml#L15

akostrikov (approved):

Issue #245, opened 2025-01-23
Do not ask info on google sheets when database is used as storage and view


zhmurov:
When USE_DATABASE_AS_STORAGE and USE_DATABASE_AS_VIEW are set, Google sheets are not used by the app. However, if GDOC_ACCOUNT_CREDENTIALS_BASE64, GDOC_SPREADSHEET_ID, etc. are not set in .env file, the app throws an exception. These parameters are read unconditionally when config is filled from_env (https://github.com/manytask/manytask/blob/main/manytask/local_config.py#L38)

PR #246, opened 2025-01-23
refactor: Improve switching between database and Google sheets

zhmurov:
Currently. there are two flags:
(1) USE_DATABASE_AS_VIEW and (2) USE_DATABASE_AS_STORAGE. The first switches viewing the results between GSheets and table that is rendered from the database. Second tells Manytask where to store the data. However, when data is stored in the database, one can not use the google sheets to view it and vise-versa. Moreover, in current implementation, one has to provide google service account credentials even when they use database instead of google sheets. This change introduces new flag 'STORAGE' that can be either GSheets (to use Google Sheets for both storage and view) and DB to use database instead. Old flags (1) and (2) are removed. Also, one does not have to specify google credentials and sheet id if they decide to use database.

TODO: Deal with link to the database page more grasefully.

Issue #245


tests/test_main.py
```diff
@@ -145,11 +144,11 @@ def test_create_app_missing_secret_key(mock_gitlab, mock_gdoc):
...
 
-def test_create_app_database_view_config(mock_env, mock_gitlab, mock_gdoc):
-    os.environ['USE_DATABASE_AS_VIEW'] = 'true'
-    app = create_app()
-    assert app.config['USE_DATABASE_AS_VIEW'] is True
-
-    os.environ['USE_DATABASE_AS_VIEW'] = 'false'
-    app = create_app()
-    assert app.config['USE_DATABASE_AS_VIEW'] is False
+#def test_create_app_database_view_config(mock_env, mock_gitlab, mock_gdoc):
```
akostrikov, line 147: remove, please if it is not needed later

manytask/main.py
```diff
@@ -157,31 +156,42 @@ def create_app(*, debug: bool | None = None, test: bool = False) -> CustomFlask:
...
             gdoc_credentials=json.loads(_gdoc_credentials_string),
-            public_worksheet_id=app.app_config.gdoc_spreadsheet_id,
-            public_scoreboard_sheet=int(app.app_config.gdoc_scoreboard_sheet),
+            public_worksheet_id=str(gdoc_spreadsheet_id),
+            public_scoreboard_sheet=gdoc_scoreboard_sheet,
             cache=cache,
         )
 
+    else:
+        raise EnvironmentError("STORAGE should be ")
```
akostrikov, line 193: What? Please, fill info about what should be and where stored

.env.example
```diff
@@ -21,12 +21,8 @@ GDOC_SPREADSHEET_ID=1cRah9NC5Nl7_NyzttC3Q5BtrnbdO6KyaG7gx5ZGusTM
...
 SHOW_ALLSCORES=true
 
-# database
-# Whether to use database as primary storage for scores and submissions
-USE_DATABASE_AS_STORAGE=false
-
-# Whether to use database for displaying scores in the web interface
-USE_DATABASE_AS_VIEW=true
+# Set primary storage for scores and submissions (GSheets or DB)
+STORAGE=GSheets
```
akostrikov, line 25: "gsheets").lower()
later in code.
Please, add as a constant and use constants in code. It is ok to use in environment as such, but lets do const.
Maybe - you want to use enum https://docs.python.org/3/library/enum.html Storage.DB or Storage.GSheet.
And probably add method Storage(from_env_str) -> Storage.DB| Storage.GSheet

akostrikov (changes requested):

tests/test_main.py
zhmurov: Removed.

manytask/main.py
zhmurov: Fixed

akostrikov:
Looks good, but needs to keep test coverage

akostrikov (approved):

PR #266, opened 2025-01-27
feat: add endpoint and modal to update student scores in the database

prawwtocol:


akostrikov (approved):
LGTM, but lets split codestyle and feature

PR #270, opened 2025-01-28
refactor: rename tester token

cin-bun:
## Description

Rename tester token

## Motivation and Context

issue - 

## Contribution Checklist

- [x] I have read and agreed on the [CONTRIBUTING](../CONTRIBUTING.md) and [CODE OF CONDUCT](../CODE_OF_CONDUCT.md) documents.
- [x] I have run linters, type checks, import sorting and tests locally.
- [x] I have added comments to code in hard-to-understand areas.
- [x] I have added 100% coverage tests for the new code (or aren't needed).
- [x] I have added docs for the new code (or aren't needed).
- [x] I have used [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) title: `tag(scope): description`.


zhmurov (dismissed):

zhmurov:
Thanks for splitting!

.env.example
```diff
@@ -2,7 +2,7 @@ FLASK_SECRET_KEY=
 
 # tokens
 REGISTRATION_SECRET=
-TESTER_TOKEN=
+MANYTASK_COURSE_TOKEN=
```
prawwtocol, line 5: why do we need the MANYTASK_ prefix if no other constans have it?

manytask/local_config.py
```diff
@@ -8,7 +8,7 @@
 class LocalConfig:
     # tokens
     registration_secret: str
-    tester_token: str
+    token: str
```
prawwtocol, line 11: why `token` here but `course_token` in other places?

.env.example
cin-bun: To make it clear that its not token from gitlab or any where else and it is exactly _course token for manytask_. For me it wasn't clear at the beginning

_Maybe be have to think about how to make clear from naming, like REGISTRATION_SECRET not good looking for me, registration where? registration what?._

manytask/local_config.py
cin-bun: agree, fixed

.env.example
prawwtocol: gitlab constants have GITLAB_ prefix, so do other constants related to external services (GDOC_)

it is implied that all other constants are related to the manytask app and thus are implied to have the MANYTASK_ prefix (so, not necessary to specify explicitly) imo

cc @akostrikov @feldsherov

cin-bun: For me looks like keeping the same style be adding MANYTASK_ to all connected with manytask, not its absence
MANYTASK_ - manytask creds
GITLAB_ - gitlab creds
GDOC_ - google doc creds

prawwtocol (approved):

Issue #237, opened 2025-01-17
Fix requiring secret word for each course


zhmurov:
To get access to the course, student must enter secret word. Before #163, the secret only needed to create user at gitlab. After #163,manytask is asking for a secret and does not create fork for the student. But one can still get to the scoreboard, without entering the secret. They will not have their fork for the public repo and will not be able to submit tasks. This is confusing and should be fixed: one should not be able to get through to the course materials until they enter the secret key.

Related to #232

cin-bun:
@zhmurov I suggest to extract checking secret code logic in decorator with param.
Something like this:
```
def requires_secret(template: str) -> Callable[..., Any]:
    def decorator(f: Callable[..., Any]):
        @wraps(f)
        def decorated(*args: Any, **kwargs: Any) -> Any:
            course: Course = current_app.course  # type: ignore
            try:
                if not secrets.compare_digest(request.form["secret"], course.registration_secret):
                    raise Exception("Invalid registration secret")
            except Exception as e:
                return render_template(
                template,
                error_message=str(e),
                course_name=course.name,
                course_favicon=course.favicon,
                base_url=course.gitlab_api.base_url,
            )
            return f(*args, **kwargs)
        return decorated
    return decorator
```

PR #271, opened 2025-01-29
feat: secret check

cin-bun:
## Description
Fix requiring secret word for each course
[#237](https://github.com/manytask/manytask/issues/237)

future tasks:
  - accord check to db user->course record
  - extract secret check from signup and create_project handlers to decorator
  - develop more clean logic for signup and logic flow 

## Motivation and Context
**Login button flow**
1. if you don't have fork
    - ask secret
    - create fork
    - redirect on main
2. you have already got fork
    - check if you have fork
    - redirect on main

### DIFF
--- a/manytask/web.py
+++ b/manytask/web.py
@@ -9,7 +9,7 @@ from flask import Blueprint, Response, current_app, redirect, render_template, r
 from flask.typing import ResponseReturnValue
 
 from . import abstract, glab
-from .auth import requires_auth, requires_ready
+from .auth import requires_auth, requires_ready, requires_secret
 from .course import Course, get_current_time
 from .database_utils import get_database_table_data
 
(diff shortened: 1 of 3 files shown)


cin-bun:
Because we can relay on existing fork as a proof of access, there is no need in storing secret in session

manytask/web.py
```diff
@@ -104,7 +107,7 @@ def get_solutions() -> ResponseReturnValue:
     course: Course = current_app.course  # type: ignore
 
     if current_app.debug:
-        if request.args.get('admin', None) in ('true', '1', 'yes', None):
+        if request.args.get("admin", None) in ("true", "1", "yes", None):
```
prawwtocol, line 110: here and later: why replace quote style?

prawwtocol: if there is a linter error, a better way would be to fix linter errors in another PR, and merge that first. 

not necessary now, keep in mind for later

cin-bun: Yes, it's linter style marks, how we've got them?

prawwtocol: :/

ok for now, I hope will be fixed in: #278

prawwtocol:
ok

prawwtocol (approved):
ok

PR #279, opened 2025-01-31
feat: Add supporting migrations to database

gagarinkomar:
Added supporting migrations using alembic library.
Added init migration.
Add param testing for DataBaseApi, because postgresql migrations don't want to work with sqlite.

TODO:
1)test corner cases
2)add tests scripts
3)try to fix problems with sqlite


manytask/migrations/env.py
```diff
@@ -0,0 +1,88 @@
...
+
+load_dotenv("../.env")
+
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# config.set_main_option("DATABASE_URL", os.environ["DATABASE_URL"])
+# config.set_main_option("SCRIPT_LOCATION", str(Path(__file__).parent / 'migrations'))
```
prawwtocol, line 17: such comments should be removed if they're not needed, because if we merge them now they will remain here forever.

manytask/database.py
```diff
@@ -39,13 +44,13 @@ def __init__(
...
         :param show_allscores: flag for showing results to all users
-        :param create_tables_if_not_exist: flag for creating database tables if they don't exist
+        :param apply_migrations: flag for applying migrations
         """
 
         self.engine = create_engine(database_url, echo=False)
 
-        if create_tables_if_not_exist:
-            self._create_tables()
+        if apply_migrations:
```
prawwtocol, line 52: I think it's better to do the following:

1. Always check for pending migrations
2. If apply_migrations=False, log a warning or raise an error

manytask/migrations/env.py
```diff
@@ -0,0 +1,88 @@
+import os
+
+from alembic import context
+from dotenv import load_dotenv
+from sqlalchemy import engine_from_config, pool
+
+from manytask.models import Base
+
+load_dotenv("../.env")
```
prawwtocol, line 9: I don't like this relative path. 


I like the following: `Path(__file__).parent.parent.parent / '.env'` but just a tiny bit better.

Why do we need to load_dotenv?

Makefile
```diff
@@ -43,4 +43,11 @@ format:
...
-	mypy --install-types --non-interactive $(ROOT_DIR) $(TESTS_DIR)
\ No newline at end of file
+	mypy --install-types --non-interactive $(ROOT_DIR) $(TESTS_DIR)
+
+makemigrations:
+	@ if [ -z "$(msg)" ]; then \
+		echo "Environment variable msg not set"; \
+		exit 1; \
+	fi
+	docker-compose -f $(DOCKER_COMPOSE_DEV) run -e EMPTY_APP=true --rm -w /app/manytask/ manytask python -m alembic revision -m "$(msg)" --autogenerate
```
prawwtocol, line 53: don't we want to simplify this command? require make dev and then just run alembic upgrade head or something similar?

prawwtocol (commented):
~I think we need to update init-db.sh to run migrations~

prawwtocol (commented):
Also, don't we want to exclude the migrations directory from our linitng rules?

manytask/alembic.ini
```diff
@@ -0,0 +1,119 @@
...
+# set to 'true' to search source files recursively
+# in each "version_locations" directory
+# new in Alembic version 1.10
+# recursive_version_locations = false
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+# sqlalchemy.url = %(DATABASE_URL)s
```
prawwtocol, line 66: please add a comment on why this is commented out

Makefile
akostrikov: Also, we need to be able to migrate code. Like `make migrate`, `make downgrade`

manytask/database.py
akostrikov: Lets log error level for now. Maybe raise error later

akostrikov (changes requested):

manytask/migrations/env.py
gagarinkomar: Changed to dotenv_values and fixed path. We need this to import database_url from .env.

akostrikov (approved):
LGTM, lets ask @prawwtocol

akostrikov (approved):

prawwtocol:
LGTM!

Issue #285, opened 2025-01-31
bug: database can be seen without secretcode


cin-bun:
/database can be seen without entering secret code and fork

Have to be fixed and check other pages and api

**scope:**
everyone
student (oAuth2 gitlab and on course)
admin

**web:**
 - _/database_              - student
 - _/solutions_              - admin
 - _/create_project_     - student
 - _/_                              - student
 - _/logout_                   - everyone
 - _/signup_                  - everyone
 - _/login_                     - everyone
 - _/not_ready_            - everyone
 
**api:**
- _/healthcheck_             - everyone
- _/report_                       - admin (tester_token)
- _/score_                        - admin (tester_token)
- _/update_config_         - admin (tester_token)
- _/update_cache_         - admin (tester_token)
- _/solutions_                  - admin (tester_token)
- _/database_                  - student
- _/database/update_     - admin

PR #287, opened 2025-01-31
fix: database secret code

cin-bun:
## Description

FIx bug when we could enter /database without secret code
Fix bug when we have to create fork if user entered secret code (but we also will create it for admin)

## Motivation and Context

https://github.com/manytask/manytask/issues/285

### DIFF
--- a/manytask/web.py
+++ b/manytask/web.py
@@ -102,6 +102,7 @@ def course_page() -> ResponseReturnValue:
 @bp.get("/solutions")
 @requires_auth
 @requires_ready
+@requires_secret(template="create_project.html")
 def get_solutions() -> ResponseReturnValue:
     course: Course = current_app.course  # type: ignore
 
@@ -304,6 +305,7 @@ def not_ready() -> ResponseReturnValue:
 @bp.get("/database")
 @requires_auth
 @requires_ready
+@requires_secret(template="create_project.html")
 def show_database() -> ResponseReturnValue:
     course: Course = current_app.course  # type: ignore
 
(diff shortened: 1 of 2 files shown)


tests/templates/base.html
prawwtocol, line 1: these files already exist in manytask/templates. can't we reuse them?

manytask/web.py
```diff
@@ -304,6 +305,7 @@ def not_ready() -> ResponseReturnValue:
 @bp.get("/database")
 @requires_auth
 @requires_ready
+@requires_secret(template="create_project.html")
```
prawwtocol, line 308: can we make this template default? I did a quick test locally and it should work

tests/test_auth.py
```diff
@@ -118,3 +120,20 @@ def __init__(self):
...
+
+    with app.test_request_context():
+        session["gitlab"] = {"oauth_access_token": "token"}
+        app.course = MagicMock()
+        app.course.registration_secret = "test_code"
+        request.form = {"secret": "test_code"}
+        app.course.gitlab_api.get_authenticated_student.return_value = None
+        app.course.gitlab_api.check_project_exists.return_value = False
+        response = test_route()
+        assert response.status_code == 302
```
prawwtocol, line 139: may be nitpicking, but maybe also check something like `response.location == "/create_project?secret=test_code"`

tests/templates/base.html
prawwtocol: p.s. in my local testing I seem to have succeeded to remove them and keep the tests functional

manytask/web.py
cin-bun: I think yes we can, but it have quite strange naming that will be fixed next in refactor auth login flow

prawwtocol: wouldn't it make sense to then consolidate this strange naming to 1 place (default arg) to make future refactoring easier?

cin-bun: yes, already making create_project default, just note that strange name will be fixed not on this pr

Issue #296, opened 2025-02-07
It is necessary to change the branch check for entries in the student's database on the course


cin-bun:
Now the verification of the need for a Secret code is based on the presence of a branch in the gitlab repository, we believe that if there is a branch, then the user already has access to the course

- It is necessary to change the check to check the entry in the database about the user on the course

PR #298, opened 2025-02-13
feat: check user on course instead check fork

cin-bun:
## Description
- new methods in db api
- check user on course before request secret

## Motivation and Context

https://github.com/manytask/manytask/issues/296


manytask/database.py
```diff
@@ -353,6 +353,38 @@ def update_task_groups_from_config(
...
+        """Sync admin flag in gitlab and db"""
+
+        with Session(self.engine) as session:
+            course = self._get(session, models.Course, name=course_name)
+            user = self._get(
+                session, models.User, username=student.username, gitlab_instance_host=course.gitlab_instance_host
+            )
+            user_on_course = self._get(session, models.UserOnCourse, user_id=user.id, course_id=course.id)
+            if student.course_admin != user_on_course.is_course_admin and student.course_admin:
+                self._update(
```
prawwtocol, line 366: do we want to handle database operation errors?

manytask/database.py
```diff
@@ -353,6 +353,38 @@ def update_task_groups_from_config(
...
+                    session=session,
+                    model=models.UserOnCourse,
+                    defaults={"is_course_admin": student.course_admin},
+                    user_id=user.id,
+                    course_id=course.id,
+                )
+                session.commit()
+            return user_on_course.is_course_admin
```
prawwtocol, line 374: we now return the old value (before the update). is it correct? don't we want to return `student.course_admin`?

tests/test_auth.py
```diff
@@ -134,7 +138,80 @@ def test_route():
...
+    def test_route():
+        return "success"
+
+    with app.test_request_context():
+        session["gitlab"] = {"oauth_access_token": "token"}
+        app.course = MagicMock()
+        app.course.registration_secret = "test_code"
+        request.form = {"secret": "test_code"}
+        app.course.gitlab_api.get_authenticated_student.side_effect = Unauthorized()
+        # app.course.storage_api.check_user_on_course.return_value = True
```
prawwtocol, line 197: please remove unused comments or leave a comment on why they're here

manytask/database.py
```diff
@@ -353,6 +353,40 @@ def update_task_groups_from_config(
...
+
+    def check_user_on_course(self, course_name: str, student: Student) -> bool:
+        """Checking tath user has been enrolled on course"""
```
prawwtocol, line 377: typo

prawwtocol (changes requested):

manytask/database.py
cin-bun: this method will be used in case where we expect that user and course already in db. That it why I think that it will be more suitable to catch exception in external methods, because we dont have any logic for it here other then just raise it

manytask/database.py
cin-bun: we dont want return `student.course_admin` because it can be different from `user_on_course.is_course_admin` and actual value is in `user_on_course.is_course_admin`

akostrikov (approved):
LGTM

prawwtocol (approved):

Issue #294, opened 2025-02-07
Make the logic of authorization and registration into two handles (/signup and /login)


cin-bun:
At the moment, this logic is divided into three handles, which complicates further work and understanding.

- It is necessary to remove the separate handle /login_finish by connecting it with /login

PR #300, opened 2025-02-14
feat: login in one handler

cin-bun:
## Description
Because of combination of two handlers login_final->login, callback URLs in GitLab setup have to be changed to /login instead of /login_final 

https://github.com/manytask/manytask/issues/294


manytask/web.py
```diff
@@ -190,54 +193,13 @@ def signup() -> ResponseReturnValue:
...
 def login() -> ResponseReturnValue:
-    """Only way to login - gitlab oauth"""
-    oauth: OAuth = current_app.oauth  # type: ignore
-
-    redirect_uri = url_for("web.login_finish", _external=True)
-
-    return oauth.gitlab.authorize_redirect(redirect_uri)
-
-
-@bp.route("/login_finish", methods=["GET", "POST"])
```
zhmurov, line 204: Looks like deleting this will affect callback URLs in GitLab setup. This should be clarified in PR description.

manytask/auth.py
```diff
@@ -23,75 +25,110 @@ def valid_session(user_session: SessionMixin) -> bool:
...
-    def decorated(*args: Any, **kwargs: Any) -> Any:
         course = current_app.course  # type: ignore
-        if not course.config:
-            return redirect(url_for("web.not_ready"))
-        return f(*args, **kwargs)
-
-    return decorated
+        oauth = current_app.oauth  # type: ignore
 
+        if "code" in request.args:
```
zhmurov, line 59: This looks odd to me, checking for the presence of an argument. Can you clarify?

cin-bun: in oAuth [autherization](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) we have required callback with code arg from auth vendor

in this decorator we handle all requests and callbacks from oauth + secret for our authorisation

zhmurov (dismissed):

manytask/auth.py
```diff
@@ -23,75 +25,110 @@ def valid_session(user_session: SessionMixin) -> bool:
...
+            # callback auth
+            try:
+                gitlab_oauth_token = oauth.gitlab.authorize_access_token()
+                student = course.gitlab_api.get_authenticated_student(gitlab_oauth_token["access_token"])
+            except Exception as e:
+                logger.error(f"Gitlab authorization failed: {e!r}")
```
prawwtocol, line 65: `logger.exception("")` (same as `logger.error("", exc_info=True)` ) includes more error information and you don't need to write !repr. 

here and later

manytask/auth.py
```diff
@@ -23,75 +25,110 @@ def valid_session(user_session: SessionMixin) -> bool:
...
         course = current_app.course  # type: ignore
-        if not course.config:
-            return redirect(url_for("web.not_ready"))
-        return f(*args, **kwargs)
-
-    return decorated
+        oauth = current_app.oauth  # type: ignore
 
+        if "code" in request.args:
+            # callback auth
```
prawwtocol, line 60: [complexity] I think it would be better to reduce cyclomatic complexity and extract some logic from require_auth. 

here, I think it would be better to extract the handle_auth logic to a separate method

manytask/auth.py
```diff
@@ -23,75 +25,110 @@ def valid_session(user_session: SessionMixin) -> bool:
...
+                student = course.gitlab_api.get_authenticated_student(session["gitlab"]["access_token"])
+                session["gitlab"].update(set_oauth_session(student))
+
+            except Exception:
+                logger.error("Failed login in gitlab, redirect to login")
+                session.pop("gitlab", None)
+                redirect_uri = url_for("web.login", _external=True)
+                return oauth.gitlab.authorize_redirect(redirect_uri)
+            # callback of secret request
+            if "secret" in request.form:
```
prawwtocol, line 83: [complexity] I think it would be better to extract handling of registration secret too

manytask/auth.py
```diff
@@ -23,75 +25,110 @@ def valid_session(user_session: SessionMixin) -> bool:
...
+                else:
+                    logger.error(f"Wrong secret user {student.username} on course {course.storage_api.course_name}")
+                    return render_template(
+                        "create_project.html",
+                        error_message="Invalid registration secret",
+                        course_name=course.name,
+                        course_favicon=course.favicon,
+                        base_url=course.gitlab_api.base_url,
+                    )
             try:
```
prawwtocol, line 95: [complexity] this could be handle_course_membership

manytask/web.py
```diff
@@ -176,7 +172,14 @@ def signup() -> ResponseReturnValue:
             raise Exception("Invalid registration secret")
         if not secrets.compare_digest(request.form["password"], request.form["password2"]):
             raise Exception("Passwords don't match")
-        _ = course.gitlab_api.register_new_user(user)
+
+        # register user in gitlab
+        gitlab_user = course.gitlab_api.register_new_user(user)
+        student = course.gitlab_api._parse_user_to_student(gitlab_user._attrs)
+        # add user->course if not in db
+        course.storage_api.sync_stored_user(student)
```
prawwtocol, line 180: 👍

tests/test_auth.py
```diff
@@ -20,6 +36,157 @@ def app():
     return app
 
 
+@pytest.fixture
+def mock_gitlab_oauth():
```
prawwtocol, line 40: maybe it's a good idea to extract common fixtures? possibly to conftest.py?

prawwtocol (changes requested):

prawwtocol (changes requested):
good

tests/test_db_api.py
```diff
@@ -842,3 +842,22 @@ def test_check_user_on_course(first_course_db_api, session):
     session.commit()
 
     assert first_course_db_api.check_user_on_course(course_name, student)
+
+
+def test_get_or_create_user(first_course_db_api, session):
```
prawwtocol, line 847: i think its better to have consistent naming, like:

test_get_or_create_user_get

or even better:

test_get_or_create_user_existing_get, test_get_or_create_user_nonexisting_create

prawwtocol (approved):

PR #309, opened 2025-02-22
fix: Redirect to signup page if auth was forgotten

### LINKED ISSUE #308: bug: Manytask forgets the user is logger into gitlab
### Is there an existing issue for this?

- [x] I have searched the existing issues, it is original

### Description of the bug

When entering the course page, the page that asks for a secret is shown with the following error:

Image

To fix, one should go to /signup page and hit the login button.

### Steps To Reproduce

I think the app forgets some attachments after not being used for a while. It happen on two different computers with two different users, but if fixed by entering via /signup page does not reproduce.

### Versions

25.1

### Additional Context

_No response_

### Terms and Conditions

- [x] I agree to follow this project's Code of Conduct.

zhmurov:
When HTTP error can be returned when information on authorized student is requested, which was not properly handled.

Fixes #308

### DIFF
--- a/manytask/web.py
+++ b/manytask/web.py
@@ -252,6 +252,13 @@ def login_finish() -> ResponseReturnValue:
 def create_project() -> ResponseReturnValue:
     course: Course = current_app.course  # type: ignore
 
+    gitlab_access_token: str = session["gitlab"]["oauth_access_token"]
+    try:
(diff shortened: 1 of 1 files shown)


zhmurov:
Does not work, investigating

zhmurov:
This does not fixes the problem, closing

PR #331, opened 2025-03-04
fix: not count/ignore removed tasks on the scoreboard and main page

gagarinkomar:
Now we save grades for removed tasks but not count/ignore they in student results.


zhmurov:
We need this in release-25 as well

zhmurov (approved):

zhmurov:
Shall I restart networks with this change on top of 2025? Will require updating DB. Or shall we do some sort of quick-and-dirty fix there? @feldsherov , what do you think?

PR #343, opened 2025-03-14
feat: keep the order of groups and tasks

gagarinkomar:
Added position fields to Task and TaskGroup models for keep the order of the initial config.
Edited queries to support the order of groups and tasks.
Added tests for it.


zhmurov (approved):

feldsherov (approved):
LGTM.

I am a bit worried about gazilion of fixtures for test configs.
Maybe once config_bundle fixture is more readable. I am not sure, though.

Issue #306, opened 2025-02-18
Remove support for GoogleSheets


zhmurov:
Google Sheets served us well, they provide good and convenient interface, allow for scripts and formulas. However, there is no feasible way to support many courses in one GSheet and if one instance serves one course, we are stuck with complicated setup procedure. Hence Google Sheets have to go as the database alternative, to clean up all the interfaces and simplify logic.

PR #353, opened 2025-03-21
refactor: Remove GoogleSheet support

zhmurov:
Google sheets do not allow for having many courses per Manytask instance. This romves the support for the Google Sheets, making the database sole storage choise.

Issue #306


feldsherov:
Awesome thing! Does it start?

Another staff which bothers me, are we sure that we can support any existing course without google sheets?
On other hand, we can use old version with google sheet support for courses, which can't be handled by new manytask.

feldsherov (approved):

zhmurov:
> Awesome thing! Does it start?
> 
> Another staff which bothers me, are we sure that we can support any existing course without google sheets? On other hand, we can use old version with google sheet support for courses, which can't be handled by new manytask.

The only feasible way to keep support for google sheets is to export data from the database there. Otherwise we are stuck with necessity of supporting very divergent logic. I think we have to make this step to be able to evolve the DB code to where we wanted. If one wants to use the old scheme, they will have to use the previous version.

zhmurov:
> Awesome thing! Does it start?

Good point. Our tests are not mature at all to rely solely on them. Will start test-instance later.

prawwtocol (approved):
lgtm

PR #354, opened 2025-03-22
refactor: Move ownership of gitlab repo urls to Course class

zhmurov:
In order to create forks for students, check if they have access and simply provide urls to the tasks, we need to know where the course group, public repo and students group are located and what is the name of the default branch. This information was placed into the GitLab API class. This was reasonable while one instance served one course: only one course was handled by this API. To be able to reuse the API for many courses, this information has to be taken out from the GitLab API class.
    
This change moves the ownership of the following data from the GitLabAPI class into the Course class:
    - gitlab_course_group - base group that encapsulates the students group and public repo
    - gitlab_course_public_repo - public repo for the course
    - gitlab_course_students_group - students group
    - gitlab_default_branch - default branch


zhmurov:
@gagarinkomar , this is somewhat related to what we discussed of Friday.

manytask/glab.py
```diff
@@ -46,10 +46,6 @@ class GitLabConfig:
 
     base_url: str
     admin_token: str
-    course_group: str
```
prawwtocol, line 49: 👍

tests/test_api.py
```diff
@@ -214,22 +214,26 @@ def __init__(self):
...
             return Student(id=TEST_USER_ID, username=TEST_USERNAME, name="", course_admin=self.course_admin)
 
         @staticmethod
-        def check_project_exists(_student):
+        def check_project_exists(_student, _course_students_group):
             return True
 
         @staticmethod
-        def _parse_user_to_student(user: dict[str, Any]):
+        def _parse_user_to_student(user: dict[str, Any], course_group, course_students_group):
```
prawwtocol, line 236: tests arent checked for typing annotataions, but it feels weird and inconsistent to not include them here. not having them, though, doesnt hurt

tests/test_glab.py
```diff
@@ -477,23 +480,23 @@ def test_get_authenticated_student_failure(mock_get, gitlab):
...
-        gitlab_api.get_authenticated_student(oauth_token)
+        gitlab_api.get_authenticated_student(oauth_token, TEST_GROUP_NAME, TEST_GROUP_STUDENT_NAME)
 
     mock_get.assert_called_once_with(f"{gitlab_api.base_url}/api/v4/user", headers=headers)
 
 
 def test_get_url_for_task_base(gitlab):
     gitlab_api, _ = gitlab
-    url = gitlab_api.get_url_for_task_base()
+    url = gitlab_api.get_url_for_task_base(TEST_GROUP_PUBLIC_NAME, "main")
```
prawwtocol, line 490: "main" is a constant that itches to be extracted for me. but it's not necessary

prawwtocol (dismissed):
lgtm in general

feldsherov (dismissed):

tests/test_api.py
zhmurov: Is there a rule when we have leading underscore and when not?

akostrikov (approved):
LGTM

PR #356, opened 2025-03-22
feat: Run tests on all pull request and when changes are pushed to any branch

zhmurov:
With more branches in the repo it is convenient to trigger tests on all PRs, not only when targeting main branch and also when changes are pushed to any branch

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,10 +1,7 @@
 name: Test
 
 on:
-  push:
-    branches: [ main ]
-  pull_request:
-    branches: [ main ]
+  on: [push, pull_request]
```
feldsherov, line 4: Should it be just on: [push, pull_request]?

feldsherov (dismissed):

.github/workflows/test.yml
zhmurov: Ops. Fixed

Issue #345, opened 2025-03-14
Move instance information out of the course object


zhmurov:
The Course object currently holds the following information:

```
class Course:
    def __init__(
        self,
        viewer_api: abstract.ViewerApi,
        storage_api: abstract.StorageApi,
        gitlab_api: glab.GitLabApi,
        solutions_api: solutions.SolutionsApi,
        registration_secret: str,
        show_allscores: bool,
        cache: BaseCache,
        manytask_version: str | None = None,
        *,
        debug: bool = False,
    ):
```

These data only partly related to the Course itself: Database and GitLab interactions should be managed by the application instance, and the information on these should be stored outside the Course.

Note that GitLab API currently holds:

```
class GitLabApi:
    def __init__(
        self,
        base_url: str,
        admin_token: str,
        course_group: str,
        course_public_repo: str,
        course_students_group: str,
        default_branch: str = "main",
        *,
        dry_run: bool = False,
    ):
```

Which is a mixture of Instance information (`base_url`, `admin_token`) and course info (`course_group`, etc.). A decision should be made if we (1) allow to store different GitLab data for different courses or (2) use one GitLab API object for Manytask instance. Option (1) is more complicated to implement, because we still rely on Manytask users being GitLab users.

zhmurov:
FYI @gagarinkomar

PR #358, opened 2025-03-28
feat: add supporting for many courses on one instance

gagarinkomar:
Added more fields to Course model to remove they from local config.
Transfered api classes from Course config to CustomFlask.
Removed requires_redy, instead of it using get_current_course function to find current user course and abort not_ready page if course not found.

It is a draft PR. Need to fix some problems with tokens and auth.
Also need to fix tests.

Not sure what solution to use api classes for instance instead of courses is better.


manytask/api.py
```diff
@@ -252,10 +241,9 @@ def report_score() -> ResponseReturnValue:
 
 
 @bp.get("/score")
-@requires_token
-@requires_ready
+@requires_tokens
 def get_score() -> ResponseReturnValue:
-    course: Course = current_app.course  # type: ignore
+    course: Course = get_current_course(request.cookies)
```
zhmurov, line 246: Is cookies correct place to store the current course?

manytask/api.py
```diff
@@ -273,7 +261,7 @@ def get_score() -> ResponseReturnValue:
 
     # ----- logic ----- #
     try:
-        group, task = course.storage_api.find_task(task_name)
+        group, task = current_app.storage_api.find_task(task_name)
```
zhmurov, line 264: Looks like task_name should be unique across all courses, which is not the case.

.env.example
```diff
@@ -1,16 +1,11 @@
 FLASK_SECRET_KEY=
 
 # tokens
-REGISTRATION_SECRET=
-MANYTASK_COURSE_TOKEN=
+MANYTASK_APP_TOKEN=55a34ae4d98bfa1ca4b29a9f8e3ab34c
```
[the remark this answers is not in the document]
akostrikov: keep in db with ability to show to admin

gagarinkomar:
TODO:
fix tests
add token autogeneration
finish requires_admin
improve frontend

manytask/config.py
```diff
@@ -12,9 +12,15 @@ class ManytaskSettingsConfig(BaseModel):
...
     course_name: str
-    gitlab_base_url: AnyUrl
-    public_repo: str
-    students_group: str
+
+    registration_secret: str
+    manytask_course_token: str
+    show_allscores: bool
+
+    gitlab_course_group: str
```
zhmurov, line 20: Are these not redundant?

zhmurov (commented):
1. Needs a lot of rebasing
2. Suggest splitting out the changes to the Course data ownership. This will make this change a lot more manageable, I think

manytask/course.py
```diff
@@ -41,23 +34,22 @@ def validate_submit_time(commit_time: datetime | None, current_time: datetime) -
 
 @dataclass
 class CourseConfig:
-    """Configuration for Course settings and APIs."""
```
zhmurov, line 44: These changes should close issue #345 . I think we should chop this into separate PR.

manytask/abstract.py
```diff
@@ -21,32 +28,38 @@ class StorageApi(ABC):
     @abstractmethod
     def get_scores(
         self,
+        course_name: str,
```
zhmurov, line 31: These can be taken out into separate PR

zhmurov (commented):
Sorry, still can't manage the volume of this PR. My suggestions are (1) move instance data from the course, (2) change the storage API to take course name and etc. as an argument and (3) remove solution API for now. These should be very easy to review and should make this change much smaller and manageable.

zhmurov:
Is this still relevant, @gagarinkomar ?

PR #359, opened 2025-03-28
fix: fixed problems with starting app

gagarinkomar:
There was a problem when starting app:

```
/usr/local/bin/python: No module named gunicorn
```


zhmurov (commented):
Suggest moving definition of env vars (line 48) up to before line 42?

zhmurov (approved):

PR #360, opened 2025-03-28
Move the envvar definition earlier

zhmurov:
In #359 the env var is defined after it is used. This moves it to before.


zhmurov:
@gagarinkomar

PR #365, opened 2025-04-01
fix: avoid division by zero

zhmurov:
If course is not initialized, or initialized with zero tasks, the
sum of the points is zero. This causes division by zero in layout,
when the percentage of completion is computed. This change conditionaly
sets the percentage to zeor is such cases.


akostrikov (approved):
LGTM

Issue #346, opened 2025-03-14
Export students First and Last names to the CSV scores table


zhmurov:
We need means to match real people with their Manytask handles while exporting grades.  This is essential for external students, who may not be in our LMS. One possible solution is to rely on students providing their names properly in GitLab and add this information to CSV export (scores table). Other solutions may work better in the long run.

zhmurov:
We may still need this, e.g. when university requests scores.

PR #368, opened 2025-04-02
feat: Show student names in the scores table and CSV export (#366)

zhmurov:
In many case it is important to match students with their names (e.g. when we don't know the name <-> nickname relation). This may be temporary solution, but it is needed currently.

Issue #346


zhmurov:
Cherry-pick to release branch.

PR #374, opened 2025-04-12
fix: Catch exception when user is deleted from gitlab

zhmurov:
If user deletes an account from gitlab, their data remains in Manytask. But when we request their name via GitLab API, exception is thrown. In this change the exception is caught and empty name is assigned to the user.


akostrikov (approved):
LGTM

PR #375, opened 2025-04-12
feat: Show the list of students names only when requested

zhmurov:
Fetching the names one by one takes time, which delays the table rendering. This adds and option to the request that enforces showing names, but does not show them by default. When needed one can add "show_full_names" oprion to the request which will result in the table with names. Later can than be exported.


akostrikov (approved):
LGTM with tests

zhmurov:
Need to work more on this:
1. Show empty column when names are not fetched
2. When enabled, it looks like it fetches the data, but doe not show the names.

zhmurov:
25.1.2 is not happening, closing this

PR #376, opened 2025-04-13
refactor: Make fields added to the Course object private

zhmurov:
Access the new fields via decorated getters.

This is a follow-up for #354


akostrikov (approved):
LGTM, but it just some syntactic sugar :)

feldsherov (approved):

PR #379, opened 2025-04-18
fix: Disable poetry cache

zhmurov:
Somehow when loaded from cache the venv appears broken. This disables caching to fix the problem.


zhmurov:
@prawwtocol , we had to do this because venv caching was broken (see here: https://github.com/manytask/manytask/actions/runs/14534654127/job/40780643460) There must be a better solution:)

PR #382, opened 2025-04-20
feat: Remove solutions API

zhmurov:
Solutions API is not used currently, so we want to remove it so it will not be slowing things down. To make sure we know how to restore its functionality, this removal should be done as a separate change.


manytask/api.py
```diff
@@ -239,7 +238,7 @@ def report_score() -> ResponseReturnValue:
 
     files = request.files.to_dict()
     if files:
```
gagarinkomar, line 240: Do we still need to save files?

PR #383, opened 2025-04-25
feat: move instance data from the course

gagarinkomar:
Moved api classes from the course.
Stored course_name instead of course in app.
Stored course config in database.
Moved some methods from the course.
Fixed tests.

This is one of the separeted PRs of https://github.com/manytask/manytask/pull/358


manytask/database.py
```diff
@@ -353,10 +366,25 @@ def sync_columns(
...
         self,
         course_name: str,
-    ) -> models.Course | None:
+    ) -> Course | None:
         try:
             with Session(self.engine) as session:
-                return self._get(session, models.Course, name=course_name)
+                course = self._get(session, models.Course, name=course_name)
+
+            return Course(
```
zhmurov, line 374: Why do we make a copy of a Course object here?

manytask/database.py
```diff
@@ -353,10 +366,25 @@ def sync_columns(
     def get_course(
         self,
         course_name: str,
-    ) -> models.Course | None:
+    ) -> Course | None:
         try:
             with Session(self.engine) as session:
-                return self._get(session, models.Course, name=course_name)
+                course = self._get(session, models.Course, name=course_name)
```
zhmurov, line 372: Course of models.Course?

manytask/abstract.py
```diff
@@ -74,6 +75,18 @@ def update_task_groups_from_config(
         deadlines_config: ManytaskDeadlinesConfig,
     ) -> None: ...
 
+    @abstractmethod
+    def update_course(
```
zhmurov, line 79: in CustomFlask we have:
```python
    def store_config(self, content: dict[str, Any]) -> None:
        manytask_config = config.ManytaskConfig(**content)

        # Update course settings
        self.storage_api.update_course(manytask_config.ui)

        # Update task groups (if necessary -- if there is an override) first
        self.storage_api.update_task_groups_from_config(manytask_config.deadlines)

        # Save deadlines to storage
        self.storage_api.sync_columns(manytask_config.deadlines)
```

This is the only place, from where ``storage_api.update_course(...)``, ``storage_api.update_task_groups_from_config(...)`` and ``storage_api.sync_columns(...)`` are called. This suggests that these three methods should be combined. Probably in a separate PR, because this requires some more work on tests.

manytask/abstract.py
```diff
@@ -74,6 +75,18 @@ def update_task_groups_from_config(
...
+        self,
+        ui_config: ManytaskUiConfig,
+    ) -> None: ...
+
+    @abstractmethod
+    def get_course(
+        self,
+        course_name: str,
```
zhmurov, line 87: The course_name is taken from storage_api when this method is called. Hence, we have repeating construct:
```python
app.storage_api.get_course(app.stogare_api.course_name)
```
Is there a reason not to get the course name inside this method and remove the argument? If so, it might be a good idea to call this method  ``get_current_course()``.

manytask/config.py
```diff
@@ -201,8 +207,9 @@ class ManytaskConfig(BaseModel):
 
     version: int  # if config exists, version is always present
 
-    settings: ManytaskSettingsConfig
```
zhmurov, line 204: This suggests that ``ManytaskSettingsConfig`` above is not needed. But you add new fields to it. Also, there are two ``ManytaskSettingsConfig`` classes, which is confusing. Do we need both of them?
Perhaps, this is a separate issue.

zhmurov (dismissed):
There are a couple of things that were revealed rather than introduced by this PR. We can file issues and merge this in its current shape to move forward.

manytask/abstract.py
zhmurov: Will resolve separately

manytask/database.py
gagarinkomar: There is a problem with class name collision.
models.Course - model class
Course(course.Course) - dataclass

I tried to simplify it. But maybe we want to rename one of the classes.

manytask/database.py
gagarinkomar: Again problem with class name collision.

zhmurov (approved):

PR #392, opened 2025-05-20
feat: add supporting many courses

gagarinkomar:
Removed course_name from app object. Also removed extra course params from .env.
Added supporting many courses.
Added admin endpoint with creating course.
Added list of courses to main page.
Slightly changed the authentication logic. Added state while gitlab login. This allows you to return to the same page where you was.
Edited tests.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,16 +1,8 @@
 FLASK_SECRET_KEY=
 
-# tokens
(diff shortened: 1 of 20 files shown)


akostrikov (approved):
LGTM

zhmurov (approved):

PR #425, opened 2025-06-17
docs: Use Diplodoc to build and deploy landing/pages

zhmurov:
1. Create landing page based on the content in the Manytask project description
2. Build existing markdown files into html using Diplodoc
3. Add action to deploy to github pages

Known problem: width of all but header-block is full-width of the browser window

Closes issue https://github.com/manytask/manytask/issues/239


zhmurov:
Does not look very nice yet: https://manytask2.org/

PR #434, opened 2025-06-24
feat: add first and last name fields to StoredUser model

KIoppert:
Now the ```users``` table also preserves the name and surname.
Relates to [#405](https://github.com/manytask/manytask/issues/405)


manytask/database.py
```diff
@@ -168,8 +168,8 @@ def sync_stored_user(
             return StoredUser(
                 username=user_on_course.user.username,
                 first_name=user_on_course.user.first_name,
-                course_admin=user_on_course.is_course_admin,
```
zhmurov: What happened here? Looks like a piece of an unrelated change

Issue #395, opened 2025-05-29
Add abstraction to the Repository Management System (RMS)


zhmurov:
Currently, the only version control provider that can be used is gitlab. Consequently, all the logic of creating project for student is tightly related to gitlab. Adding another layer of abstraction will allow for using manytask with other providers, e.g. Sourcecraft when it will have an API.

This issue should be addressed after authorization is separated from gitlab (see issue #295)

PR #435, opened 2025-06-24
refactor: Create RmsUser object and use it

zhmurov:
Currently, the student object represent both RMS and logged user. This can not be the case when the authorization and RMS logic wil be separated. This change creates new object to store RMS user and uses it thoughout the code. With that, the separation between authorization and repository management becomes more visible.

Issue #395


akostrikov (approved):
LGTM as steps in right direction

KIoppert (dismissed):

PR #436, opened 2025-06-24
refactor: Do not use Student in storage API

zhmurov:
Student object is ambiguous and used all over with different purpose. This change removes this object from the storage API making it more clear on what is passed there as arguments.

### DIFF
--- a/manytask/api.py
+++ b/manytask/api.py
@@ -291,7 +291,7 @@ def get_score() -> ResponseReturnValue:
             )
         else:
             assert False, "unreachable"
-        student_scores = course.storage_api.get_scores(student.username)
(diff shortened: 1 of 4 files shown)


DmitryMelentsov:
LGTM, but the merge conflict needs to be resolved

manytask/api.py
```diff
@@ -196,6 +196,7 @@ def report_score(course_name: str) -> ResponseReturnValue:
         logger.info(f"Got score=None; set max score for {task.name} of {task.score}")
 
     student = _get_student(app.gitlab_api, user_id, username)
+    first_name, last_name = student.name.split()  # TODO: come up with how to separate names
```
zhmurov, line 199: These indicate where storage API is not good and should be updated/fixed

PR #439, opened 2025-06-27
feat: add course selector

DmitryMelentsov:


tests/test_web.py
```diff
@@ -114,6 +114,14 @@ def __init__(self):
         def get_scores_update_timestamp(_course_name):
             return datetime.now(tz=ZoneInfo("UTC"))
 
+        @staticmethod
```
zhmurov, line 117: Looks like this should not be here

manytask/templates/layout.html
```diff
@@ -65,7 +65,7 @@
                     </li>
                     {% if show_allscores %}
                     <li class="nav-item">
-                        <a class="nav-link px-3 {% if active_page == 'database' %}active{% endif %}" 
+                        <a class="nav-link px-3 {% if active_page == 'database' %}active{% endif %}"
```
zhmurov, line 68: A change with no change.

zhmurov (approved):
A couple very minor issues and this is good to go in

PR #440, opened 2025-06-28
feat: Do not create user when store_score(..) and get/sync_stored_user(..) are called

zhmurov:
Currently, in the guts of store_score(..), get_stored_user(..) and sync_stored_user(..) functions, the user is searched for and, if not found, is created. This should not be the case, because
- store_score(..) is user in API to set score(s) for task(s) and one should not try to set scores for non-existing user.
- get_stored_user(..) should not change anything - just return the user info
- sync_stored_user(..) should not create user in the system, only user on course

Also added tests that check that these functions throw an exception if called on non-existing user.

The function names are also confusing, which is in part why these functions are misused, see #449


KIoppert (dismissed):
LGTM

manytask/database.py
```diff
@@ -755,17 +755,14 @@ def _apply_migrations(self, database_url: str) -> None:
     def _get_or_create_user_on_course(
         self,
         session: Session,
-        student: Student,
+        username: str,
```
gagarinkomar, line 758: Don't we want to change the signature in all other methods too?(except for the user creation method)
Now pass username and then change to some id?

tests/test_db_api.py
gagarinkomar, line 1: Do we have a test to verify that the user is not created in the modified methods?(I know that everything is not perfect with the tests, so you we can skip it.)

zhmurov: Done. Also made a (draft) follow-up, where I added the exception handling.

Issue #447, opened 2025-06-30
Do not store GitLab instance host in User and Course


zhmurov:
GitLab instance should be the same for Manytask instance, hence there is no need to save it in DB

PR #448, opened 2025-06-30
feat: Do not save GitLab instance url in Users and Courses DB models

zhmurov:
GitLab instance is the same for the entire Manytask app, so it should not be saved for all users and courses in the database.

Issue https://github.com/manytask/manytask/issues/447


manytask/models.py
```diff
@@ -91,9 +91,8 @@ class User(Base):
     username: Mapped[str]
     first_name: Mapped[str]
     last_name: Mapped[str]
-    gitlab_instance_host: Mapped[str]
 
-    __table_args__ = (UniqueConstraint("username", "gitlab_instance_host", name="_username_gitlab_instance_uc"),)
+    __table_args__ = (UniqueConstraint("username"),)
```
gagarinkomar, line 95: Instead, we can simply write like this:
```
username: Mapped[str]= mapped_column(unique=True)
first_name: Mapped[str]
first_name: Mapped[str]
```
We used
```
__table_args__ = (UniqueConstraint("username", "gitlab_instance_host", name="_username_gitlab_instance_uc"),)
```
because we needed the uniqueness of the pair. Now we need uniqueness of only one field.

zhmurov: Done

PR #450, opened 2025-07-03
feat: added validation, as well as additional information

KIoppert:
1) Improved understanding of what needs to be entered in the fields
2) Added validation
3) Added CSRF protection (only for ```create_course and``` ```create_project``` endpoints)
https://test.manytask2.org/admin/courses/new


manytask/main.py
```diff
@@ -9,12 +9,15 @@
...
 from dotenv import load_dotenv
 from flask import Flask
+from flask_wtf import CSRFProtect
 from werkzeug.middleware.proxy_fix import ProxyFix
 
 from . import abstract, config, course, database, glab, local_config
 
 load_dotenv("../.env")  # take environment variables from .env.
 
+csrf = CSRFProtect()
```
gagarinkomar, line 19: Will it work if we make this object a CustomFlask field?

manytask/templates/create_course.html
gagarinkomar, line 1: I'm not very good at frontend but It looks good. It's better for someone else to analyse javascript, because I don't fully understand what's going on in it.

Do we have basic checks for the length of the secret, the name of the course, etc...? I think we should discuss it at the meeting.

gagarinkomar (commented):
I haven't had to work with CSRF in flask yet. But running through code, it looks like this changes should protect us from CSRF. 

Did you have the opportunity to exploit this vulnerability and make sure everything works?

manytask/main.py
KIoppert: Yes, indeed, corrected

zhmurov (approved):

PR #452, opened 2025-07-05
feat: Add exception handling for _get_or_create_user_on_course(..)

zhmurov:
When _get_or_create_user_on_course(..) is called an exception can be thrown if the user does not exist in database. This exception should be properly handled (mainly in the API, where there are no other checks for the username).


manytask/api.py
```diff
@@ -207,13 +207,19 @@ def report_score(course_name: str) -> ResponseReturnValue:
...
+            course.course_name,
+            student.username,
+            app.rms_api.get_url_for_repo(student.username, course.gitlab_course_students_group),
+            task.name,
+            update_function,
+        )
+    except Exception as e:
+        return (
+            f"There was an exception on updating score for user {student.username}: {str(e)}",
+            HTTPStatus.NOT_FOUND,
```
akostrikov, line 221: NOT_FOUND may not be suitable for all

akostrikov (changes requested):

zhmurov:
I think this is deprecated now.

PR #457, opened 2025-07-07
refactor: Consolidate identical code in one function in REST API

zhmurov:
There are identical expressions in two API functions, which are handled
separately. This consolidates them into one function. The tests are also
refactored to reflect the change.

### DIFF
--- a/manytask/api.py
+++ b/manytask/api.py
@@ -11,7 +11,9 @@ import yaml
 from flask import Blueprint, abort, current_app, jsonify, request, session
 from flask.typing import ResponseReturnValue
 
+from manytask.abstract import StorageApi
 from manytask.database import TaskDisabledError
+from manytask.glab import GitLabApi
 
 from .auth import requires_auth, requires_ready
 from .config import ManytaskGroupConfig, ManytaskTaskConfig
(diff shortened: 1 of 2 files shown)


manytask/api.py
```diff
@@ -112,13 +116,23 @@ def _validate_and_extract_params(form_data: dict[str, Any]) -> tuple[str, int |
...
-        check_deadline = form_data["check_deadline"] is True or form_data["check_deadline"] == "True"
+    """Get student by user_id or username."""
+    try:
+        if username:
+            student = gitlab_api.get_student_by_username(username)
+        elif user_id:
+            student = gitlab_api.get_student(user_id)
+        else:
+            assert False, "unreachable"
+    except Exception:
```
KIoppert, line 127: Will be better if we caught specific exceptions : ```GitLabApiException```, etc.

KIoppert, line 127: I think it's worth casting ```form_data["user_id"]``` to ```int``` inside ```try...except``` block and catch ```ValueError```. It will be more safer

KIoppert (changes requested):

manytask/api.py
```diff
@@ -112,13 +116,23 @@ def _validate_and_extract_params(form_data: dict[str, Any]) -> tuple[str, int |
...
-    if "check_deadline" in form_data:
-        check_deadline = form_data["check_deadline"] is True or form_data["check_deadline"] == "True"
+    """Get student by user_id or username."""
+    try:
+        if username:
+            student = gitlab_api.get_student_by_username(username)
+        elif user_id:
+            student = gitlab_api.get_student(user_id)
+        else:
+            assert False, "unreachable"
```
elyaishere, line 126: `student = gitlab_api.get_student_by_username(username) if username else gitlab_api.get_student(user_id)`

manytask/api.py
zhmurov: Note that there may also be exception from gitlab, but these should be handled in gitlab api itself and this is out of scope here.

KIoppert: Understood, I'm sorry

zhmurov: Changed to throwing GitLabApiException

zhmurov: Added try-except

manytask/api.py
zhmurov: Changed to:
```
if "both defined":
   BAD_REQUEST
elif "user_id":

KIoppert (approved):
LGTM

manytask/api.py
zhmurov: Will do a follow-up if needed. Resolving for now

PR #458, opened 2025-07-07
feat: add first and last name in score table

KIoppert:
Now, the DB looks like this:
image

```Name``` is replaced by two columns: ```First Name``` and ```Last Name```. Search is working on any column.

And some UI updates (you can see this on screenshot), to my mind now it looks better, but I'm ready to listen criticism

### DIFF
--- a/manytask/database_utils.py
+++ b/manytask/database_utils.py
@@ -26,11 +26,13 @@ def get_database_table_data(app: CustomFlask, course_name: str) -> dict[str, Any
     for username, student_scores in all_scores.items():
         total_score = sum(student_scores.values())
         large_count = sum(1 for task in large_tasks if student_scores.get(task, 0) > 0)
-        student_name = app.gitlab_api.get_student_by_username(username).name
+        stored_user = storage_api.get_stored_user(course_name, username)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: manytask/database_utils.py [hunk 1]
```diff
@@ -26,11 +26,13 @@ def get_database_table_data(app: CustomFlask, course_name: str) -> dict[str, Any
     for username, student_scores in all_scores.items():
         total_score = sum(student_scores.values())
         large_count = sum(1 for task in large_tasks if student_scores.get(task, 0) > 0)
-        student_name = app.gitlab_api.get_student_by_username(username).name
+        stored_user = storage_api.get_stored_user(course_name, username)
```
zhmurov, line 29: The 'StorageApi.get_all_scores(..)' is only used here. So we can join first and last names when the scores are extracted there. This should save us some calls to database.

Suggest making and issue and planning follow-up for this, since it will require some work on tests.

zhmurov (approved):
LGTM

Suggest making an issue from my comment to come back to it later.

manytask/database_utils.py
KIoppert: #459

PR #461, opened 2025-07-09
perf: optimize calls to db when table is rendering

KIoppert:
This PR improve speed of rendering table at twice. Even visually I can see strong difference.
Some benchmark tests (if I wrote their correctly :) ):
```test_db (NOW)``` - **old** realisation
```test_db (0004_...)``` - **new** realisation
When we have ```1000``` users on course:
image
When we have ```500``` users:
image
When we have ```100``` users:
image

But I had to change a signature of method ```get_all_scores```, now it is ```get_all_scores_with_names```. In this method I make new ```dict``` for names, where key is ```username```, but we already have ```dict``` with key ```username```(```all_scores```). And maybe will be good if we combine them, but in my case tests is refactoring very quickly and code is readable

### DIFF
--- a/manytask/database.py
+++ b/manytask/database.py
@@ -177,7 +177,7 @@ class DataBaseApi(StorageApi):
         """
 
         with Session(self.engine) as session:
-            all_users = self._get_all_users(session, course_name)
+            all_users = self._get_all_users_on_course(session, course_name)
 
         scores_and_names: dict[str, tuple[dict[str, int], tuple[str, str]]] = {}
         for user in all_users:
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: manytask/database.py [hunk 1]
```diff
@@ -170,7 +170,9 @@ def sync_stored_user(
                 course_admin=user_on_course.is_course_admin,
             )
 
-    def get_all_scores(self, course_name: str) -> dict[str, dict[str, int]]:
+    def get_all_scores_with_names(
+        self, course_name: str
+    ) -> Tuple[dict[str, dict[str, int]], dict[str, tuple[str, str]]]:
```
zhmurov, line 175: Would it be better to return 'dict[str, tuple[str, str, dict[str, int]]'? This way we will not be searching for the first / last name in the dict for each user.

KIoppert:
Added loading spinner, while DB table is rendering (UI update)

zhmurov (dismissed):

zhmurov (approved):

PR #463, opened 2025-07-12
fix: final score calculation with reported score

elyaishere:


zhmurov (approved):
LGTM, but needs proper commit message

PR #466, opened 2025-07-18
feat: automate initial course setup

DmitryMelentsov:


manytask/glab.py
```diff
@@ -88,48 +104,83 @@ def _get_project_by_name(self, project_name: str) -> gitlab.v4.objects.Project:
...
+                    "visibility": "public",
+                    "shared_runners_enabled": True,
+                    "auto_devops_enabled": False,
+                    "initialize_with_readme": True,
+                    "default_branch": default_branch,
+                }
+            )
+        except gitlab.GitlabError:
+            try:
+                self._get_project_by_name(f"{group.full_path}/{course_public_repo}")
```
zhmurov, line 151: We should check existence first, and than create project.  Using try-except this was is not conventional and confusing (at least for me).

manytask/glab.py
```diff
@@ -88,48 +104,83 @@ def _get_project_by_name(self, project_name: str) -> gitlab.v4.objects.Project:
...
+                    "path": course_students_group,
+                    "parent_id": group.id,
+                    "visibility": "private",
+                    "lfs_enabled": True,
+                    "shared_runners_enabled": True,
+                }
+            )
+        except gitlab.GitlabError:
+            try:
+                self._get_group_by_name(f"{group.full_path}/{course_students_group}")
```
zhmurov, line 178: Same here: first check existence, only than - try to create.

manytask/abstract.py
```diff
@@ -168,6 +168,12 @@ def register_new_user(
         password: str,
     ) -> None: ...
 
+    @abstractmethod
+    def create_course_group(
```
zhmurov, line 172: This just creates group of projects on gitlab. Suggest calling it "create_group"

manytask/glab.py
akostrikov: Agree

PR #477, opened 2025-07-27
feat: remove unnecessary session refreshes

gagarinkomar:
I'm not sure, but are there any situations where this is needed now?


akostrikov (approved):
LGTM

PR #486, opened 2025-07-28
refactor: Refactor requires_auth function 

zhmurov:
get_authenticated_student(..) is called when both getting student info
and checking if the user is logged in. This add check_authenticated_student(..)
method that is used when we need to check if the user is logged in.

### DIFF
--- a/manytask/auth.py
+++ b/manytask/auth.py
@@ -89,7 +89,7 @@ def get_authenticate_student(oauth: OAuth, app: CustomFlask) -> Student:
     return student
 
 
-def __redirect_to_login() -> Callable[..., Any]:
+def __redirect_to_login() -> Response:
     session.pop("gitlab", None)
     return redirect(url_for("root.signup"))
 


manytask/auth.py
```diff
@@ -80,18 +81,17 @@ def handle_oauth_callback(oauth: OAuth, app: CustomFlask) -> Response:
...
+    student = app.gitlab_api.get_authenticated_student(session["gitlab"]["access_token"])
+    session["gitlab"].update(set_oauth_session(student))
+    return student
 
-    except Exception:
-        logger.error("Failed login in gitlab, redirect to login", exc_info=True)
-        session.pop("gitlab", None)
-        return redirect(url_for("root.signup"))
+
+def __redirect_to_login() -> Callable[..., Any]:
```
KIoppert, line 92: Should this return `Response`, not `Callable[..., Any]`?

KIoppert (approved):

manytask/auth.py
zhmurov: It looks like this is separate issue

KIoppert: It's strange, my `mypy` doesn't swear at it.

zhmurov: Ok. I have two virtual environments, one has mypy version 1.15.0, and this is where I get this error. With mypy 1.17.0 everything seems to be fine.

zhmurov:
@gagarinkomar , can you double-check please?

Issue #414, opened 2025-06-06
Store gitlab/RMS login in the database


zhmurov:
After authentication will be separated from code management, the user name (login) for Repository Management System will not be the same as the authenticated user loging. Hence it can not be extracted from the session and should be saved somewhere. The validation of user identity should also be added.

zhmurov:
store RMS ID only

PR #494, opened 2025-07-30
feat: Store RMS ID in the database

zhmurov:
To be able to match user with RMS user, we need to save their RMS ID in the database.

TODO: Use the RMS ID when updating score with API

Issue #414


manytask/migrations/versions/5a4c2adf497b_store_rms_id_in_user_table.py
```diff
@@ -0,0 +1,30 @@
...
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('rms_id', sa.Integer(), nullable=True))
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('users', 'rms_id')
+    # ### end Alembic commands ###
```
KIoppert, line 30: Need to regenerate migration, because `rms_id` now is unique

zhmurov: Tried to merge two migration files. Needs double-checking

PR #497, opened 2025-08-01
refactor: Return bool instead of throwing when checking authenticated RMS user

zhmurov:
Checking if user logged into RMS should not throw and exception: this is a routine check and nothing is wrong when user is not authenticated. Hence this function should return boolean instead of throwing an exception.

Changed the name of the function accordingly.

### DIFF
--- a/manytask/glab.py
+++ b/manytask/glab.py
@@ -8,6 +8,7 @@ import gitlab
 import gitlab.const
 import gitlab.v4.objects
 import requests
+from urllib.error import HTTPError
 
 from .abstract import RmsApi, RmsUser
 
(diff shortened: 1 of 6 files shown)


manytask/glab.py
```diff
@@ -3,6 +3,7 @@
 import logging
 from dataclasses import dataclass
 from typing import Any
+from urllib.error import HTTPError
```
zhmurov, line 6: requests.exception

manytask/glab.py
```diff
@@ -267,13 +268,18 @@ def get_rms_user_by_username(
...
+    def check_user_authenticated_in_rms(
         self,
         oauth_token: str,
-    ) -> None:
+    ) -> bool:
         headers = {"Authorization": "Bearer " + oauth_token}
         response = requests.get(f"{self.base_url}/api/v4/user", headers=headers)
-        response.raise_for_status()
+        try:
+            response.raise_for_status()
```
KIoppert, line 278: `self.get_authenticated_rms_user(oauth_token)`

I'm not really sure, but how about getting rid of the copypaste a bit? Or move common part to a separate method?

tests/test_api.py
```diff
@@ -491,6 +491,7 @@ class Course:
                 "version": 1.5,
                 "username": TEST_USERNAME,
                 "user_id": TEST_USER_ID,
+                "access_token": 123,
```
KIoppert, line 494: nitpick: string

PR #538, opened 2025-08-14
refactor: Store instance not course admin in StoredUser

zhmurov:
Stored user contains username, first and last name, rms_id, these represent user, not user on course.


manytask/database.py
```diff
@@ -660,7 +661,7 @@ def get_all_users(self) -> list[StoredUser]:
                     first_name=user.first_name,
                     last_name=user.last_name,
                     rms_id=user.rms_id,
-                    course_admin=user.is_instance_admin,
```
zhmurov, line 663: This is the only place where returned value is read from StoredUser and it is obviously instance admin here.

KIoppert (dismissed):

Issue #525, opened 2025-08-05
bug: for `initial_instance_admin` not defined `rms_id`


KIoppert:
After #494 zero admin can't create project for some course

zhmurov:
Can be fixed manually once per instance. Note that first and last names should be updated as well.

PR #543, opened 2025-08-20
fix: fix for "add grade config" PR

KIoppert:
Fix problems with migration in #428


manytask/database.py
```diff
@@ -82,7 +82,7 @@ def __init__(
                 session,
                 models.User,
                 username=config.instance_admin_username,
-                defaults={"first_name": "Instance", "last_name": "Admin", "is_instance_admin": True},
+                defaults={"first_name": "Instance", "last_name": "Admin", "is_instance_admin": True, "rms_id": -1},
```
KIoppert, line 85: Need more successful decision. Relates to #525

PR #555, opened 2025-08-23
fix: Set RMS ID for initial instance admin when app is initialized

### LINKED ISSUE #525: bug: for `initial_instance_admin` not defined `rms_id`

zhmurov:
When initial instance admin is created, their RMS ID is set to -1. This causes an error when this user is identified later. Moreover, there is no means to update the RMS ID from admin panel to fix this error. This fixes the issue by getting RMS ID when instance is initialized.

Fixes #525


tests/test_main.py
```diff
@@ -95,6 +95,12 @@ def mock_project_list_search(**kwargs):
 
         gitlab_instance.projects.list.side_effect = mock_project_list_search
 
+        # Mock GitLab RMS user for instance admin
+        mock_rms_user = MagicMock()
+        mock_rms_user.username = os.getenv("INITIAL_INSTANCE_ADMIN", "instance_admin")
+        mock_rms_user.id = 12345
+        gitlab_instance.get_rms_user_by_username.return_value = mock_rms_user
+
```
KIoppert, line 103: You need this instead of what you have
```py
mock_student = MagicMock()
mock_student._attrs = {"id": 0, "username": "", "name": "Test User"}

def mock_get_student(**kwargs):
    username = kwargs.get("username", "")
    mock_student._attrs["username"] = username
    return [mock_student]

gitlab_instance.users.list.side_effect = mock_get_student
```

zhmurov: It worked, thanks!

manytask/main.py
```diff
@@ -136,10 +136,13 @@ def _database_storage_setup(app: CustomFlask) -> abstract.StorageApi:
     if instance_admin_username is None:
         raise EnvironmentError("Unable to find INITIAL_INSTANCE_ADMIN env")
 
+    rms_user = rms_api.get_rms_user_by_username(instance_admin_username)
+
```
KIoppert, line 140: We need if..else block for `app.debug`

KIoppert (changes requested):

KIoppert (dismissed):

zhmurov:
Don't want to step on @gagarinkomar s heals with this one. So the priority is low.

prawwtocol (changes requested):

manytask/database.py
```diff
@@ -85,7 +86,11 @@ def __init__(
                 models.User,
                 username=config.instance_admin_username,
                 defaults={"is_instance_admin": True},
```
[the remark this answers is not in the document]
zhmurov: Тут dafaults же не поможет, получается? Нужно явно где-то поменять его?

prawwtocol (dismissed):

zhmurov:
Better fix is already in the code. the fixes for tests are now in PR #929.

PR #557, opened 2025-08-23
feat: add link to admin panel on not-ready page

KIoppert:
image


zhmurov (approved):
check_admin(...) may be useful elsewhere as well

PR #569, opened 2025-08-27
feature: rms_api class implementation for SourceCraft

Siegmeyer1:


manytask/sourcecraft.py
```diff
@@ -0,0 +1,223 @@
...
+        elif response.status_code == HTTPStatus.NOT_FOUND:
+            return False
+        else:
+            response.raise_for_status()
+            return False
+
+    def _create_student_repo(self, course_name: str, login: str) -> None:
+        """This method triggers SourceCraft CI, so creation is actually async and
+        returned json is an operation which caller can poll to check if creation is finished.
+        """
```
Siegmeyer1, line 113: This actually can be changed to smth like "clone public template repo -> create new repo in SC -> push template code to new repo". This will make the method synchronous but will require presence of "public template repo" for course

manytask/abstract.py
```diff
@@ -196,25 +196,25 @@ def create_students_group(
...
     ) -> bool: ...
 
     @abstractmethod
     def create_project(
         self,
         rms_user: RmsUser,
-        course_students_group: str,
-        course_public_repo: str,
+        destination: str,
+        template: str,
```
zhmurov, line 207: This is not a template, but repository. By design, templates are something that is updated rarely, but this repo is constantly updated with new tasks as the course progresses. Students then update their forks to get new tasks. In our terminology, this is called public repository. We can discuss the naming, but i would not prioritize this. And removing this from the PR will make the footprint of the later much smaller.

zhmurov (commented):
If we create a mock for auth api, will we be able to test-try this?

manytask/sourcecraft.py
Siegmeyer1: Want to keep it as is for now. This way we keep SourceCraft-specific logic inside SourceCraft

PR #576, opened 2025-09-01
feat: Support interpolated scores

domwst:
# Problem

Currently in manytask config there is an option of soft deadlines:
https://github.com/manytask/manytask/blob/main/.manytask.example.yml#L15-L20
however it is not used anywhere in the manytask codebase and implemented purely on the checker's side, which is strange and limits usages of manytask with custom checkers. This PR implements server-side scoring for soft deadlines.

## Proposed configuration

We already have steps in deadline config:

```yaml
    - group: group_0
      start: 2000-01-01 18:00
      steps:
        0.5: 2000-03-01 18:00
      end: 2000-05-01 23:59
```

Let's reuse them for interpolated score:

```yaml
    - group: group_0
      start: 2000-01-01 18:00
      steps:
        1.0: 14d
        0.5: 21d
      end: 2000-05-01 23:59
```

In the interpolated case configuration above will give full score for submissions in the first 14 days, then during the next 7 days score will be linearly reduced to a half and will remain like that until the end

### DIFF
--- a/manytask/config.py
+++ b/manytask/config.py
@@ -213,13 +213,6 @@ class ManytaskFinalGradeConfig(BaseModel):
     grades_order: list[int] = Field(default_factory=list)
 
     @model_validator(mode="after")
-    def check_grade_names_initialization(self) -> ManytaskFinalGradeConfig:
-        if len(self.grades.keys()) == 0:
-            raise ValueError("No grades defined")
-
-        return self
-
-    @model_validator(mode="after")
     def populate_grades_order(self) -> ManytaskFinalGradeConfig:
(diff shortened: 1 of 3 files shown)


manytask/course.py
```diff
@@ -83,6 +89,7 @@ def __init__(
         self.registration_secret = config.registration_secret
         self.token = config.token
         self.show_allscores = config.show_allscores
+        self.deadlines_type = config.deadlines_type
```
domwst, line 92: What is rationale behind having `CourseConfig` and `Course` with the exact same set of fields?

elyaishere (approved):

PR #587, opened 2025-09-04
fix: allow table visibility without final grade config

dmasloff:


manytask/config.py
```diff
@@ -209,6 +202,10 @@ def evaluate(self, scores: dict[str, Any]) -> Optional[int]:
             if ManytaskFinalGradeConfig.evaluate_grade(self.grades[grade], scores):
                 return grade
 
+        # shortcut for courses that do not use builtin grading system
+        if len(self.grades_order) == 0:
+            return 0
```
zhmurov, line 207: Ideally, we should not show the column in this case. But column with zeros should be fine for now.

PR #589, opened 2025-09-05
perf: add cache for static resources

KIoppert:
This increases loading speed for all pages ~0.1-0.2 seconds

### DIFF
--- a/manytask/main.py
+++ b/manytask/main.py
@@ -8,12 +8,14 @@ from typing import Any
 import yaml
 from authlib.integrations.flask_client import OAuth
 from dotenv import load_dotenv
-from flask import Flask, Response, request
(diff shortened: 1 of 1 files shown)


manytask/main.py
```diff
@@ -101,6 +101,12 @@ def create_app(*, debug: bool | None = None, test: bool = False) -> CustomFlask:
 
     logger.info("Init success")
 
+    @app.after_request
```
prawwtocol, line 104: OK but could've also been 

`app.config['SEND_FILE_MAX_AGE_DEFAULT'] = 3600` afaik

KIoppert: done

prawwtocol (approved):

PR #597, opened 2025-09-08
feat: bonus column in the course table

KIoppert:
Relates to #591 
This add ability to set bonus scores for the students.

### DIFF
--- a/tests/test_db_api.py
+++ b/tests/test_db_api.py
@@ -622,7 +622,7 @@ def test_store_bonus_score(db_api_with_initialized_first_course, session):
     assert session.query(User).count() == 1
     assert session.query(UserOnCourse).count() == 0
 
-    db_api_with_initialized_first_course.create_user_if_not_exist(
+    db_api_with_initialized_first_course.update_or_create_user(
         TEST_USERNAME, TEST_FIRST_NAME, TEST_LAST_NAME, TEST_RMS_ID
     )
 


manytask/models.py
```diff
@@ -206,6 +206,7 @@ class UserOnCourse(Base):
     course_id: Mapped[int] = mapped_column(ForeignKey(Course.id))
     join_date: Mapped[datetime] = mapped_column(server_default=func.now())
     is_course_admin: Mapped[bool] = mapped_column(default=False)
+    bonus_score: Mapped[int] = mapped_column(default="0", server_default="0")
```
prawwtocol, line 210: why `"0"` and not `0`? why not `nullable=False`?

manytask/database.py
```diff
@@ -348,7 +356,7 @@ def store_score(self, course_name: str, username: str, task_name: str, update_fn
                 grade.last_submit_date = datetime.now(timezone.utc)
 
                 session.commit()
-                logger.info(f"Setting score = {new_score}")
+                logger.info(f"Setting score = {new_score} for {user_on_course.user.username} to {task_name}")
```
prawwtocol, line 359: i dont like this log string because
1. it is ambiguous -- what is "to" and "for"?
2. user.id is better than username for debugging

so i suggest something like

```
logger.info(
    "Setting score to %d for user_id=%s (username=%s) on task=%s",
    new_score,
    user_on_course.user.id,
    user_on_course.user.username,
    task_name,
)
```

I'm using parametrized logging because if the log level becomes disabled parametrized logging does 0 actions while f strings would still be computed

manytask/database.py
```diff
@@ -337,6 +338,13 @@ def store_score(self, course_name: str, username: str, task_name: str, update_fn
                 user_on_course = self._get_or_create_user_on_course(session, username, course)
                 session.commit()
 
+                if task_name == "Bonus":
```
prawwtocol, line 341: i really really don't like this string hardcoding here and in the jinja template. also, I don't like that the bonus score lives in `user_on_course` and regular scores in the same function live in `grade`

what are we even trying to accomplish in this PR? why not just create a task called "Bonus task" in the tasks DB table? and then edit it's scores without having to add ANY logic?

prawwtocol (changes requested):

manytask/database.py
KIoppert: This is really important point. Open new issue #603

manytask/database.py
KIoppert: We are trying to add column bonus for extra scores (e.g. for mistakes in tasks, contributing to the course etc.)

It is a good idea to create extra task so as not to change the logic. 
Now task group `Bonus group` and task `bonus_score` are created in each course, when config updates.
Task group is disabled to hide this group in Assigments page.

PR #601, opened 2025-09-09
feat: Student repo links

domwst:
On the `database` page with scores we now display the links to the students' repositories for course or instance admins:
image

### DIFF
--- a/manytask/api.py
+++ b/manytask/api.py
@@ -323,9 +323,7 @@ def get_database(course_name: str) -> ResponseReturnValue:
         abort(HTTPStatus.NOT_FOUND, "Course not found")
 
     rms_user = app.rms_api.get_rms_user_by_id(session["gitlab"]["user_id"])
-    # More round trips to the God of round trips!
-    is_course_admin = storage_api.check_if_course_admin(course.course_name, rms_user.username) or \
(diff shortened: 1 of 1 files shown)


manytask/api.py
```diff
@@ -317,8 +317,16 @@ def update_cache(course_name: str) -> ResponseReturnValue:
 def get_database(course_name: str) -> ResponseReturnValue:
     app: CustomFlask = current_app  # type: ignore
 
+    storage_api = app.storage_api
+    course = app.storage_api.get_course(course_name)
+    if course is None:
+        abort(HTTPStatus.NOT_FOUND, "Course not found")
+
+    rms_user = app.rms_api.get_rms_user_by_id(session["gitlab"]["user_id"])
```
prawwtocol, line 325: i understand that this was copied but wouldn't the user be able to change their session username and become an admin this way? 

if true, this is a huge security risk

prawwtocol (changes requested):

prawwtocol (commented):
other than that, LGTM

manytask/api.py
prawwtocol: ok, good to know

PR #602, opened 2025-09-10
feat: Database api authorization via course token

domwst:


zhmurov:
An explanation on why we need this shall be useful

PR #616, opened 2025-09-19
refactor: optimize repo existance check

dmasloff:
related to #549

### DIFF
--- a/manytask/glab.py
+++ b/manytask/glab.py
@@ -152,9 +152,11 @@ class GitLabApi(RmsApi, AuthApi):
             logger.info(f"Project exists project_name={project_name} group={project_group}")
             return True
 
-        logger.info(f"Project does not match the expected pattern:\n"
-                    f"got project candidate path={projects[0].path_with_namespace}\n"
-                    f"awaited project_name={project_name} group={project_group}")
+        logger.info(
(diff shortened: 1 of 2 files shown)


KIoppert (approved):

manytask/glab.py
```diff
@@ -141,16 +141,22 @@ def check_project_exists(self, project_name: str, project_group: str) -> bool:
...
-            if project.path_with_namespace == gitlab_project_path:
-                logger.info(f"Project exists project_name={project_name} group={project_group}")
-                return True
+        if len(projects) == 0:
+            logger.info(f"Project does not exist project_name={project_name} group={project_group}")
+            return False
 
-        logger.info(f"Project does not exist project_name={project_name} group={project_group}")
+        logger.debug(f"Found project candidate path={projects[0].path_with_namespace}")
+        if projects[0].path_with_namespace == gitlab_project_path:
```
zhmurov, line 151: This relies on the condition above to not fail, maybe use if-else there?

zhmurov (approved):

manytask/glab.py
dmasloff: It is more usual to remove additional branching if can be done, though I am open to any codestyle suggestions

PR #704, opened 2025-10-30
feat: Add is_solved field into Grade table

### LINKED ISSUE #703: Store large homeworks results in the database
Currently, LHW count is re-computed based on the score threshold criteria every time the course database table is generated. This disallows one to disable counting the number of large homework when 'doreshka' is activated, does not allow to override LHW result manually. To overcome these issues, we should store if the LHW is solved in the database.

zhmurov:
In order to track if the LHW was solved before the deadline, we need to save solved/not solved state in the database.

Fixes #703


manytask/utils/database.py
```diff
@@ -30,15 +30,24 @@ def get_database_table_data(app: CustomFlask, course: Course, include_admin_data
...
-        large_count = sum(1 for task in large_tasks if student_scores.get(task[0], 0) >= task[1])
+        total_score = 0
+        for score_solved in student_scores.values():
+            total_score = total_score + score_solved[0]
+
+        large_count = 0
+        for large_task in large_tasks:
+            if student_scores.get(large_task[0], (0, None))[1]:
+                large_count += 1
+
```
KIoppert: Suggest using generators:

```py
total_score = sum(score_solved[0] for score_solved in student_scores.values())

large_count = sum(1 for large_task in large_tasks if  student_scores.get(large_task[0], (0, None))[1])
```

zhmurov: Done

KIoppert (approved):

PR #739, opened 2025-12-01
fix: resolve session profile mismatch security vulnerability

D1sney:
## Problem
The application used session["profile"]["username"] for authorization checks,
but this profile was not synchronized with the currently authenticated user
session["gitlab"]["username"]. This created a critical security vulnerability:

- If User A logged in, then User B logged in using the same browser, User B
  would receive User A's permissions
- This allowed unauthorized access to admin features and other users' data

## Root Cause
The profile was set once during signup and never updated when users switched
accounts. The code relied on this stale profile data for authorization decisions.

## Solution
1. Add profile validation in signup_finish() endpoint to detect and clear
   mismatched profiles
2. Replace all uses of session["profile"]["username"] with
   session["gitlab"]["username"] in authorization checks

## Changes
- manytask/web.py: 
  - Add profile mismatch detection in signup_finish (lines 233-242)
  - Fix authorization checks in course_page, show_database, create_project
- manytask/api.py:
  - Fix authorization in update_database and update_comment endpoints
- manytask/utils/flask.py:
  - Fix user identification in get_courses helper

## Security Impact
This vulnerability allowed privilege escalation and unauthorized data access.
The fix ensures that authorization checks always use the current user's identity.

## Testing
Manually verified that switching between admin and non-admin accounts now
correctly enforces permissions.

### DIFF
--- a/manytask/api.py
+++ b/manytask/api.py
@@ -351,7 +351,11 @@ def get_database(course_name: str, auth_method: AuthMethod) -> ResponseReturnVal
     if auth_method == AuthMethod.SESSION:
         rms_user = app.rms_api.get_rms_user_by_id(session["gitlab"]["user_id"])
         is_course_admin = storage_api.check_if_course_admin(course.course_name, rms_user.username)
-        logger.info("[DEBUG] API /database auth_method=SESSION, username=%s, is_course_admin=%s", rms_user.username, is_course_admin)
+        logger.info(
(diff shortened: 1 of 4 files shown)


gagarinkomar:
@zhmurov , это устаревший ПР. Сейчас подобной проблемы в коде не вижу, она была решена в https://github.com/manytask/manytask/pull/854 . ПР можно закрывать.

PR #808, opened 2026-02-03
feat: implement RmsApi for SourceCraft

Siegmeyer1:


zhmurov (approved):
We can merge this based on the fact that it does not touch existing functionality. But it does not seem clear to me how to enable sourcecraft backend& Should there be an if somewhere?

PR #811, opened 2026-02-05
feat: implement AuthApi for SourceCraft

Siegmeyer1:


zhmurov (commented):
To make the full story we should:
- Investigate and option to use GitLab with Yandex ID
- Add mock for authentication so that we will not have strange logic when GitLab is only used for authentication

We should plan to reconsider envvar names in future releases.

manytask/main.py
```diff
@@ -52,29 +52,48 @@ def create_app(*, debug: bool | None = None, test: bool = False) -> CustomFlask:
...
+        app.rms_api = sourcecraft.SourceCraftApi(
+            sourcecraft.SourceCraftConfig(
+                base_url=app.app_config.sourcecraft_url,
+                api_url=app.app_config.sourcecraft_api_url,
+                admin_token=app.app_config.sourcecraft_admin_token,
+                org_slug=app.app_config.sourcecraft_org_slug,
+            ),
+            app.storage_api,
+        )
+    else:
```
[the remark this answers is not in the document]
Siegmeyer1: Done

manytask/main.py
```diff
@@ -52,29 +52,58 @@ def create_app(*, debug: bool | None = None, test: bool = False) -> CustomFlask:
...
+                org_slug=app.app_config.sourcecraft_org_slug,
+            ),
+            app.storage_api,
+        )
+
     elif rms == "mock":
+        app.oauth = _authenticate(
+            OAuth(app), app.app_config.gitlab_url, app.app_config.gitlab_client_id, app.app_config.gitlab_client_secret
+        )
+        app.auth_api = glab.GitLabApi(  # TODO: mock auth api
```
zhmurov: Can you do this mock so we don't accumulate copy-paste dept?

zhmurov: The idea for the mock was:
1. Remove code replication from tests
2. Allow running an app locally, without RMS

But having a mock should also allow to see more clearly what is going to happen when authenticated user is not the same thing as RMS user.

PR #817, opened 2026-02-09
fix: recalculate grades on update_config

### LINKED ISSUE #812: Re-calculate grades when update_config(...) is called
Currently, when the grading config is updated, we do not recalculate students grade. When new tasks are issued, the total possible number of points can go up. Hence the percent of completion for the students should go down, which can be followed by the decrease in grade. It looks like if the student remains quiet in this case, they can keep their grade, because it is never recomputed.

D1sney:
Summary
- Recalculate all students' grades after config update using existing scoring logic.
- Extend test mocks to support grade recalculation in update_config.

Testing
- uv run pytest tests/test_api.py::test_update_config_success -v
- uv run pytest tests/test_api.py -v

Fixes #812

### DIFF
--- a/manytask/utils/database.py
+++ b/manytask/utils/database.py
@@ -95,7 +95,6 @@ def get_database_table_data(
         else:
             effective_grade = storage_api.calculate_and_save_grade(course_name, username, row)
 
-
         row["grade"] = effective_grade
 
         # Add override indicator using already loaded data


manytask/api.py
```diff
@@ -492,6 +492,53 @@ def update_config(course_name: str) -> ResponseReturnValue:
         logger.exception("Error while updating config for course=%s", course_name, exc_info=True)
         return f"Invalid config for course={course_name}", HTTPStatus.BAD_REQUEST
 
+    # Recalculate grades after config update (best-effort)
+    try:
```
zhmurov: there is a bunch of very similar code in https://github.com/manytask/manytask/blob/main/manytask/utils/database.py , we should try to partly reuse it

PR #875, opened 2026-03-09
feat: Improve error reporting for /report API (fixes #702)

### LINKED ISSUE #702: Improve error reporting for /report API handle
Currently the report handle does not tell at all what went wrong, e.g. when token is invalid or API endpoint is wrong

Siegmeyer1:
manytask plugin (checker/plugins/manytask.py): Before sending the report request, log request URL, task name, username, score, and a token hint (length only; if length > 16, also first 4 characters). Never log the full token. On HTTP errors (4xx/5xx), include the request URL in the exception message.

tester (checker/tester.py): When starting the report pipeline, log how many stages it has. When reporting fails, print a short summary for each stage (name, status: failed/skipped/ok) and, for failed stages, the stage output/error text line by line.

pipeline (checker/pipeline.py): When a plugin raises PluginExecutionFailed, store both error.output and error.message in the stage result’s output so that the tester can show full error details when reporting fails.

### DIFF
--- a/checker/checker/tester.py
+++ b/checker/checker/tester.py
@@ -157,6 +157,15 @@ class Tester:
             pipeline_conf = self.testing_config.report_pipeline
         return PipelineRunner(pipeline_conf, self.plugins, verbose=self.verbose)
 
+    def _print_report_failure_details(self, result: PipelineResult) -> None:
+        """Print per-stage status and output when report pipeline fails."""
+        for stage in result.stage_results:
+            status = "FAILED" if stage.failed else ("skipped" if stage.skipped else "ok")
+            print_info(f"  stage '{stage.name}': {status}", color="red" if stage.failed else "grey")
+            if stage.output:
+                for line in stage.output.splitlines():
+                    print_info(f"    {line}", color="grey")
+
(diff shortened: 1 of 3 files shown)


checker/checker/tester.py
```diff
@@ -250,13 +259,17 @@ def run(
...
+                    f"Reporting <{task.name}> task tests ({len(report_pipeline)} stage(s) in report pipeline):",
+                    color="pink",
+                )
                 if report:
                     task_report_result: PipelineResult = report_pipeline.run(context, dry_run=self.dry_run)
                     if task_report_result:
                         print_info("->Reporting succeeded")
                     else:
                         print_info("->Reporting failed")
+                        self._print_report_failure_details(task_report_result)
```
elyaishere: может воспользоваться? [str](https://github.com/manytask/manytask/blob/6b6a760a18c3da55d9051d4571ddf4eeb6152513/checker/checker/pipeline.py#L48)(task_report_result) и добавить output в печать PipelineStageResult

elyaishere (changes requested):

zhmurov:
superseded by #931

PR #905, opened 2026-04-02
feat: add a Checker plugin that scores by the result of a custom script

zhmurov:
If the problem can not be solved exactly it can be graded by the quality of the solution by a custom number. E.g. the traveling salesman problem solution can be judged by the length of the final route - shorter is better. This adds a plugin that runs a user-provided script and extracts the score from the output. Note that in current state it is very easy for the student to print the wanted result and exit so it should be made more secure.

### DIFF
--- a/checker/checker/plugins/run_script_with_score.py
+++ b/checker/checker/plugins/run_script_with_score.py
@@ -1,0 +1,130 @@
+from __future__ import annotations
+
+import re
+from typing import Optional, Union
+
+from pydantic import Field
+
+from checker.exceptions import PluginExecutionFailed
+from checker.plugins import PluginABC, PluginOutput
+from checker.plugins.scripts import RunScriptPlugin
+
+
+class RunScriptWithScorePlugin(RunScriptPlugin):
+    """Plugin for running a custom script that reports a score.
+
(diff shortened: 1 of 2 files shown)


checker/checker/plugins/run_script_with_score.py
```diff
@@ -0,0 +1,130 @@
...
+            default=r"(?:^|[^\d.])(\d+(?:\.\d+)?)(?:[^\d.]|$)",
+            description=(
+                "Regex with one capture group that matches the score value in the script output. "
+                "The *last* match in the output is used. The matched value must be in [0, 1]."
+            ),
+        )
+
+    def _run(self, args: Args, *, verbose: bool = False) -> PluginOutput:  # type: ignore[override]
+        script_cmd: str | list[str] = args.script
+        if args.allow_failures or args.report_score:
```
Dmi4er4, line 65: `report_score=True` здесь автоматически проглатывает ненулевой exit code даже при `allow_failures=False`. Упавший scorer, успевший вывести число, в итоге дает валидную оценку. Кажись надо оборачивать команду только при `allow_failures`, иначе дефолт флага не работает

PR #917, opened 2026-04-08
feat: Add line number column to students table

### LINKED ISSUE #915: Add a column with the line number to the list of students table

dotsystemsdevs:
## Summary

Adds a new '#' column at the beginning of the students/users table in the namespace panel to display line numbers. This improves readability and makes it easier to reference specific students.

## Changes

- Added '#' header column in the users table
- Uses Jinja2 `loop.counter` to automatically number rows starting from 1
- Line numbers are displayed in the first column before User ID

## Benefits

- Easier to reference students in discussions/communications
- Better readability in tables with many students
- Common UX pattern in data tables

## Affected File

- `manytask/templates/namespace_panel.html`

Closes #915

### DIFF
--- a/manytask/templates/namespace_panel.html
+++ b/manytask/templates/namespace_panel.html
@@ -50,6 +50,7 @@
                     <table class="table table-striped table-hover">
                         <thead>
                             <tr>
+                                <th>#</th>
                                 <th>User ID</th>
                                 <th>Username</th>
                                 <th>RMS ID (GitLab)</th>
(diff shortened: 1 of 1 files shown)


zhmurov:
Sorry, this is the wrong table. We meant one in the database.html template. Number might be useful in this table as well. though.

prawwtocol (changes requested):
Sorry, this is the wrong table. We meant one in the database.html template. Number might be useful in this table as well. though.

PR #920, opened 2026-04-10
fix: return empty string when group path is requested for group id -1 in sourcecraft

zhmurov:
Namespaces are not yet supported in Sourcecraft, but it safer to treat them as they are in the same group ("") as compared to throwing every time someone tries to list available namespaces


zhmurov:
It fixed the problem on src.manytask.org

PR #931, opened 2026-04-29
feat: better logging for task report method

Siegmeyer1:


manytask/api.py
```diff
@@ -281,21 +303,21 @@ def _validate_and_extract_params(
...
-            abort(HTTPStatus.NOT_FOUND, f"There is no RMS user with id {user_id}")
+            abort(HTTPStatus.NOT_FOUND, f"There is no RMS user with id={user_id}")
     elif "username" in form_data:
         username = form_data["username"]
         try:
             rms_user = rms_api.get_rms_user_by_username(username)
             logger.info("Found RMS user by username=%s: %s", sanitize_log_data(username), rms_user.username)
         except RmsApiException:
-            abort(HTTPStatus.NOT_FOUND, f"There is no RMS user with username {username}")
+            abort(HTTPStatus.NOT_FOUND, f"There is no RMS user with username='{username}'")
```
zhmurov: Do we need to sanitize username?

manytask/api.py
```diff
@@ -307,7 +329,10 @@ def _validate_and_extract_params(
         course, group, task = storage_api.find_task(course_name, task_name)
     except (KeyError, TaskDisabledError):
         logger.warning("Task not found or disabled: %s", sanitize_log_data(task_name))
-        abort(HTTPStatus.NOT_FOUND, f"There is no task with name `{task_name}` (or it is closed for submission)")
+        abort(
+            HTTPStatus.NOT_FOUND,
+            f"Task '{task_name}' not found in course '{course_name}' (or it is closed for submission)",
```
zhmurov: do we need to sanitize?

manytask/api.py
```diff
@@ -307,7 +329,10 @@ def _validate_and_extract_params(
         course, group, task = storage_api.find_task(course_name, task_name)
     except (KeyError, TaskDisabledError):
```
zhmurov, line 330: we can split by exception here if we do another iteration with the PR, but it is good as it is

zhmurov (dismissed):

manytask/api.py
```diff
@@ -281,21 +303,21 @@ def _validate_and_extract_params(
     """Validate and extract parameters from form data."""
 
     if "user_id" in form_data and "username" in form_data:
-        abort(HTTPStatus.BAD_REQUEST, "Both `user_id` or `username` were provided, use only one")
+        abort(HTTPStatus.BAD_REQUEST, "Both `user_id` and `username` were provided, use only one")
     elif "user_id" in form_data:
         user_id = form_data["user_id"]
         try:
             rms_user = rms_api.get_rms_user_by_id(user_id)
```
zhmurov, line 310: User ID here is, in fact, RMS ID. Which now should be string. We do need to make it more clear what is expected in the API request, probably by changing parameter names. Now I suggest adding WRONG_RMS_ID string parameter to the test constants: this problem is clearly not related to the current PR.

zhmurov (approved):

zhmurov:
@elyaishere this replaces #875

PR #942, opened 2026-05-05
monorepo: part 1

prawwtocol:
Part 1 of monorepo restructure: only git mv, no fixes yet. Build/import errors will be fixed in follow-up branches.


zhmurov (approved):

zhmurov (commented):
Shall we keep VERSION in the root?

PR #945, opened 2026-05-05
monorepo: part 4

prawwtocol:
Part 1 of monorepo restructure: only git mv, no fixes yet.
Build/import errors will be fixed in follow-up branches.


zhmurov (commented):
Not needed if the version file remains in the root

zhmurov (approved):

PR #947, opened 2026-05-05
monorepo: part 6

prawwtocol:
Part 1 of monorepo restructure: only git mv, no fixes yet.
Build/import errors will be fixed in follow-up branches.


zhmurov (approved):
Do we need a Makefile in the root that can run all the checks (perhaps we should wait until we have integration tests though).

PR #952, opened 2026-05-05
monorepo: part 7

prawwtocol:
common yml config


common.example.yml
```diff
@@ -0,0 +1,77 @@
+# Единый плоский конфиг приватного курса (пример)
```
zhmurov, line 1:
```suggestion
# Unified flat course config example
```

zhmurov (approved):
Where did the .manytas.example.yml go?

zhmurov (commented):
Here: https://github.com/manytask/manytask/blob/fix-monorepo-part-7/manytask/.manytask.example.yml . Now this is duplicated

PR #968, opened 2026-05-14
fix: Fix local dev scripts

zhmurov:
- Fix relative path to the docker compose file
- Re-create the PAT if its value is lost since we can only retreat this value when token is created 
- Do not start docs container


prawwtocol (approved):
LGTM!

PR #988, opened 2026-06-08
ops(compose): use shared reverse-proxy and add postgres to production

Dmi4er4:
The VM hosts a shared nginx-proxy + acme-companion stack at /srv/nginx-proxy/ that fronts multiple manytask stands (manytask.org, manytask2.org, az-*, src-*, etc). The previous production compose defined its own nginx-proxy/acme-companion services with the same container_name, which clashed on every deploy.

Drop the proxy services from this compose and consume the shared one via the external nginx-manytask-net network. Reattach a postgres service so the production stack is self-contained; the data volume manytask_postgres_data is external and preserved across redeploys.

manytask depends_on postgres healthcheck, so the app no longer races the DB on cold start.


zhmurov:
I think this is the step in the right direction. Questions are:
1. Do we want to take out docs container and move it into docs folder?
2. Do we have instructions on how to set everything up on fresh server?

In my head, we need to have something like this:
- In root folder we have auxiliary docker compose file that runs nginx and acme
- In manytask folder we have docker compose that runs web app with database
- In docs - one for docs
- ... ?

Thoughts?

docs/docker-compose.production.yml
```diff
@@ -0,0 +1,32 @@
...
+
+services:
+  docs:
+    build:
+      context: .
+    image: manytask-docs
+    container_name: manytask-docs
+    restart: always
+    environment:
+      VIRTUAL_HOST: ${DOCS_HOST}
```
zhmurov: We need .env for this, don't we?

zhmurov (commented):
Nice! I think this is the modularity we need.

docs/docker-compose.production.yml
Dmi4er4: Good catch and it's exactly one variable, ${DOCS_HOST}. Important nuance, that's a compose-time interpolation, not a container runtime var, so a service-level `env_file:` wouldn't feed it. env_file populates the container's env after it's built, ${DOCS_HOST} is substituted by compose before that. It has to come from the project env-file.
 Both APP_HOST and DOCS_HOST already live in manytask/.env, and the deploy passes --env-file manytask/.env for the whole umbrella, so all three stacks interpolate from that one file:
`docker compose --env-file manytask/.env -f compose/docker-compose.production.yml up -d`
 Verified on the stand, VIRTUAL_HOST resolves to app.manytask2.org / manytask2.org in docker compose config.
 For a standalone `docker compose -f docs/docker-compose.production.yml up` you'd pass the same --env-file ../manytask/.env. Will make guard to fail loudly instead of silently  defaulting to blank when someone forgets

zhmurov (approved):

PR #1018, opened 2026-07-27
feat: clone+push from template on project creation to preserve history

Siegmeyer1:


zhmurov:
Will this break existing courses?

PR #1023, opened 2026-08-02
fix: remove check_if_instance_admin(..) wrapper (fixes admin routes 500)

Ch0p1k3:
## Motivation

Same class of bug as #1022 (just merged), same fix pattern.

`manytask/utils/flask.py` used to define a `check_if_instance_admin(app)` helper whose name shadowed the underlying `storage_api.check_if_instance_admin(username)` method. In `auth.py::requires_instance_or_namespace_admin` the caller intended the storage-API method but grabbed the imported wrapper instead and passed a `username: str`. The wrapper then did `app.debug` on that string:

```
AttributeError: 'str' object has no attribute 'debug'
```

which surfaces as **HTTP 500** on every `@requires_instance_or_namespace_admin` route (`/instance_admin/courses/new`, `/instance_admin/courses/<name>/edit`, etc.) whenever `app.debug is False` (i.e. all real deployments).

#1022 fixed the sibling `check_if_course_admin` collision by deleting the wrapper altogether. This PR does the same for `check_if_instance_admin`.

## Changes

- `utils/flask.py` — delete `check_if_instance_admin(app)` (13 lines)
- `auth.py::requires_instance_or_namespace_admin` — drop the wrapper import, call `app.storage_api.check_if_instance_admin(username)` directly
- `web.py` — remove wrapper from top-level and local imports; three call sites (`index`, `not_ready`, `instance_admin_index`) now call `storage_api` directly. Debug bypass preserved inline where relevant:
  ```python
  is_instance_admin = True if app.debug else app.storage_api.check_if_instance_admin(username)
  ```

Net: **+7 / -20**.

## Backward compatibility

Bug fix only — no public API changes, no config changes, no schema changes. Behaviour of the removed wrapper is preserved exactly at every call site (debug bypass included).

## Tests

No new tests. The existing `tests/test_auth.py` and `tests/test_web.py` still pass (verified locally). No test referenced the deleted wrapper — only `storage_api.check_if_instance_admin`, which is unchanged.

## Verification

- `uv run --extra dev ruff check manytask` — clean
- `uv run --extra dev mypy manytask` — clean (21 files, no issues)
- `grep -rn check_if_instance_admin manytask/manytask/` — every remaining reference is either a `def` in `abstract.py` / `database.py` (the storage-API method itself) or a `storage_api.check_if_instance_admin(...)` call site

### DIFF
--- a/manytask/manytask/auth.py
+++ b/manytask/manytask/auth.py
@@ -297,7 +297,7 @@ def requires_instance_or_namespace_admin(f: Callable[..., Any]) -> Callable[...,
     @requires_auth
     @wraps(f)
     def decorated(*args: Any, **kwargs: Any) -> Any:
-        from .utils.flask import check_if_instance_admin, check_if_user_has_namespaces_to_admin
+        from .utils.flask import check_if_user_has_namespaces_to_admin
 
         app: CustomFlask = current_app  # type: ignore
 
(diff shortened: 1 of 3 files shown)


Ch0p1k3:
Closing — will let maintainers address the naming conflict in their own preferred way (similar to #1022 for the sibling function). Happy to reopen if useful.

PR #1034, opened 2026-08-05
fix: remove link to GitLab account from the drop-down menu

### LINKED ISSUE #1021: fix: remove "account settings"
### Is there an existing planned feature for this?

- [x] I have searched the existing issues, it is original

### Type of feature

`fix` - small bug fix

### Current behavior

Button "account settings" for SourceCraft leads to non-existent page on SourceCraft

### Suggested solution

Remove button entirely. Manytask do not create nor manage SourceCraft accounts so there is no point to have a link to them on the main page

### Additional context

_No response_

### Terms and Conditions

- [x] I agree to follow this project's Code of Conduct.

zhmurov:
This link is not universal and works only for gitlab. It has nothing to do with manytask itself and user can find their profile in gitlab interface.

Closes #1021

### DIFF
--- a/manytask/manytask/web.py
+++ b/manytask/manytask/web.py
@@ -74,7 +74,6 @@ def index() -> ResponseReturnValue:
         can_create_courses=can_create_courses,
         is_instance_admin=is_instance_admin,
         username=username,
-        gitlab_url=app.rms_api.base_url,
     )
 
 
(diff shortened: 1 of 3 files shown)


manytask/manytask/web.py
```diff
@@ -152,7 +151,6 @@ def course_page(course_name: str) -> ResponseReturnValue:
         course_name=course.course_name,
         course_status=course.status,
         app=app,
-        gitlab_url=app.rms_api.base_url,
```
Siegmeyer1, line 155: are you sure that it is used for profile link only?

zhmurov: Yes. There is a `task_base_url` above for linking tasks in students repos

Siegmeyer1 (approved):

PR #1046, opened 2026-08-16
fix: Adapt cpp checker to new test environment

domwst:


checker/checker/plugins/cpp_forbidden.py
```diff
@@ -32,18 +86,35 @@ def _run(self, args: Args, *, verbose: bool = False) -> PluginOutput:  # type: i
...
         forbidden: list[str] = []
-        for f in args.forbidden:
-            forbidden += ["-f", f]
-        for f in args.forbidden_files:
-            forbidden += ["-ff", f]
+        # Workaround for firejail limit of 128 cli arguments.
+        # Some problems have ~50 forbidden features,
+        # which results in ~100 cli arguments from forbidden
+        # features list alone if passed separately.
+        if args.forbidden:
```
domwst, line 93: Note that this breaks compatibility with older versions of `check_forbidden` tool. If anyone else is currently using it we have to do some sort of version dispatch here

zhmurov: can we return meaningful error if the number of args is more than 1 (or even better, allow using both many args and ; separator?
I did not find a course that uses this plugin though, so breaking is fine here

zhmurov (approved):

checker/checker/plugins/cpp_forbidden.py
domwst: Currently if `-f` flag is supplied more than once, the last occurrence takes precedence – this is the intended behavior of LLVM's cli args parser. I can make `check_forbidden` fail explicitly in such case or support several occurrences of these flags, but I'm not sure whether it would be any useful as this plugin is the only thing that uses the tool.

zhmurov: I think adding a clear message if there are more than 1 arg is enough. We can merge as is and keep an eye on how this is used accross the courses (didn't have time to check all of them but it seems like only C++ courses are using it)

PR #1057, opened 2026-08-26
Feat add course admin management to course edit page

zhmurov:
The table shows everyone who can administer a course in one table, regardless of
where the access comes from: instance admins, namespace admins, program
managers and course admins. Also added a management for course admins directly
from this page so one can add/remove teachers there.


zhmurov:
Screenshot 2026-08-26 at 21 03 28

PR #1060, opened 2026-08-31
feat(gitlab): grant students reporter access to the course public repo

ObjatieGroba:
On instances where the course public repo lives in a private GitLab, students otherwise have no access to it once their personal fork is created. Explicit REPORTER membership on the public project lets them read task statements and the shared CI config referenced via ci_config_path. The grant is idempotent: a GitlabCreateError (already a member) is swallowed, mirroring the neighboring DEVELOPER grant on the forked project.


zhmurov (approved):

ObjatieGroba:
Should it be config flagged?

zhmurov:
> Should it be config flagged?

I think it is fine without it: it does not hurt to add student as a reported even if the public repo is public/internal

PR #1062, opened 2026-08-31
draft: feat: store submission history and recalculate scores on config update

ObjatieGroba:
Submissions (raw score, submit time, flags, commit sha) are now persisted per grade; task scores are recomputed from the full history when the course config changes, so deadline/max-score edits apply retroactively; commit_sha is informational (taken from CI_COMMIT_SHA) and enables reviewing the exact submitted commit; groundwork for attempt-count-based scoring.

Внимание! Нельзя переключать во время семестра - старые результаты могут пропасть при наличии новых посылок

### DIFF
--- a/checker/checker/configs/checker.py
+++ b/checker/checker/configs/checker.py
@@ -113,6 +113,8 @@ class CheckerTestingConfig(CustomBaseModel):
 
     search_plugins: list[str] = Field(default_factory=list)
 
+    report_on_failure: bool = False
+
     global_pipeline: list[PipelineStageConfig] = Field(default_factory=list)
     tasks_pipeline: list[PipelineStageConfig] = Field(default_factory=list)
     report_pipeline: list[PipelineStageConfig] = Field(default_factory=list)
--- a/checker/checker/plugins/manytask.py
+++ b/checker/checker/plugins/manytask.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import os
 from datetime import datetime
 from pathlib import Path
 from typing import IO, Any, Optional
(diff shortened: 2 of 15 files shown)


zhmurov:
Question: Can we restore the student's code from the commit hash alone? Or we need to set up a repo where all the attempts are pushed?
Suggestions:
1. Split the part that saves the submission into separate PR. This way we are definitely not breaking score computing.
2. What about failed attempts? Currently the score is sent only when tests are passing. Though this depends on how one configures checker/grader/whatever is used.

ObjatieGroba:
Thanks for the review.                                                                                                                                                                                   
                                                                                                                                                                                                               
Splitting submissions storage from score recalculation. I'd rather keep them in one PR: the two are tightly coupled in code. Recalculation reads its input from the stored submissions, and when run_penalty is enabled the live score is derived from that same history, so a separate "storage only" PR would be dead code with no consumer and no way to test it meaningfully.                         
                                                                                                                                                                                                               
Failed attempts. Good catch, I had missed that. Fixed: the checker gets a new testing.report_on_failure flag (default false). When set, the report pipeline also runs for a failed task pipeline, so the attempt reaches manytask with score 0 and counts towards the penalty. The task itself is still marked as failed.                                                                                         
                                                                                                                                                                                                               
Restoring code from the commit hash. I think storing only the commit sha is enough; we don't need to store the submission itself. That already covers most cases, since the commit lives in the student's repo. To guard against force pushes making the commit unreachable, the CI job could additionally create a branch or a tag per submission in the student's repo. That is a separate, optional step and doesn't affect this PR.

PR #1064, opened 2026-09-02
Feat add course admin management to course edit page

zhmurov:
Before namespaces, the only option to give an admin access to the course was by giving instance admin status. To assign (and re-assign) roles, we need to have good tools for changing them. This adds a table to the course editing page that allows to see who has access to the course and by what scope and allows changing the course admin status.

Makes use of #1063


zhmurov:
Screenshot 2026-09-09 at 09 13 06

PR #1076, opened 2026-09-08
feat: do not show the secret info on course editing page by default

zhmurov:
Course secret and token are exposed on the course editing page. This may lead to someone accidently showing these information when demonstrating their screen. This hides the characters by default. A toggle is added to reveal the info when needed.


zhmurov:
Screenshot 2026-09-08 at 08 26 40

zhmurov:
check if one can copy

zhmurov:
> check if one can copy

works

PR #1092, opened 2026-09-14
feat(checker): skip_unchanged_tasks option to skip untouched tasks in…

ObjatieGroba:
Pulling newly published tasks from the public repo makes them show up as changed for changes_detection; they get graded against the template solution, fail, and with report_on_failure are reported as failed submissions. With testing.skip_unchanged_tasks set to the name of the task parameter listing student-editable files (e.g. allow_change), `checker grade` compares those files byte-for-byte with what the export publishes (.template file/folder, template comments rendered, or the reference file itself) and skips tasks that are identical: not run, not reported.


zhmurov:
If there are changes in the public repo that are in the task student already submitted, the checks will be triggered on this task I think. Is this expected and we want to re-grade in this case?

ObjatieGroba:
> If there are changes in the public repo that are in the task student already submitted, the checks will be triggered on this task I think. Is this expected and we want to re-grade in this case?

That shouldn't happen. If already-submitted tasks weren't re-graded before, this change won't cause it either. It's only an extra pre-launch check that complements the main one rather than replacing it, so no re-grading is expected.

PR #1096, opened 2026-09-20
fix: don't 500 on RMS failure in create_project

Ch0p1k3:
## Problem

`create_project` catches only `gitlab.GitlabError`:

```python
except gitlab.GitlabError as ex:
    logger.error("Project creation failed: %s", ex.error_message)
    return render_template(app.signup_template, error_message=ex.error_message, ...)
```

Every RMS backend (`glab`, `sourcecraft`, `mock_rms`) raises `RmsApiException` instead, and `sourcecraft.py` alone raises it from 14 places. So on SourceCraft any backend failure escapes the handler and Flask returns a bare 500.

Hit in production on a course with ~460 students: the cloud organization ran out of its access-bindings quota, so repo creation started failing for everyone at once.

```
POST https://api.sourcecraft.tech/orgs/<org>/repos -> 429 Too Many Requests
...
  File "/app/manytask/web.py", line 393, in create_project
    app.rms_api.create_project(rms_user, course.gitlab_course_students_group, course.gitlab_course_public_repo)
  File "/app/manytask/sourcecraft.py", line 349, in create_project
  File "/app/manytask/sourcecraft.py", line 168, in _create_repo
    raise RmsApiException(f"Failed to create repo: {response.json()}")
manytask.abstract.RmsApiException: Failed to create repo: {'error_code': 'ResourceExhausted',
  'message': '[Cloud] Resource exhausted: Access bindings quota of 2000 ... has been exceeded'}
```

From the student's side this is a broken page with no hint of what happened, and support gets "manytask throws 500, and I never got a repo invite" — two symptoms of one cause.

## Change

1. **Catch `RmsApiException`.** This is the backend-agnostic path, so it covers SourceCraft, GitLab and any future RMS. The raw message is backend-internal (API payloads, cloud resource ids, request ids), so it is logged and the user gets a readable sentence instead.
2. **Render `create_project.html` on error**, not `app.signup_template`. The user submitted the enrollment form; returning the signup page is confusing on its own. The `gitlab.GitlabError` branch keeps surfacing its own message, just on the right page now.
3. **Pass the full template context on every error path.** The `GET`, and `Invalid secret` branches already passed `course_name` / `course_favicon` / `base_url`; the CSRF and failure branches did not, so the favicon and the "Back to courses" link broke exactly when the user needed them. Extracted a small `render_create_project()` closure so the four call sites cannot drift apart again.

## Tests

Two regression tests in `tests/test_web.py`. Both fail on `main` and pass with this change:

- `test_create_project_renders_error_instead_of_500_when_rms_fails` - `RmsApiException` gives 200 with the enrollment form and a readable message; the backend-internal detail is not in the body.
- `test_create_project_still_reports_gitlab_errors` - the GitLab path still shows its own message, now on the create_project page.

`make lint` is clean (`ruff format`, `ruff check`, `mypy`). Locally the DB suites error out on testcontainers/Docker, unrelated to this change.

## Not covered here

`glab.create_project` raises bare `RuntimeError` when the course group is missing, which still produces a 500. That looks like a course misconfiguration rather than a user-facing failure, so I left it alone - happy to fold it in if you would rather it be handled too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


manytask/manytask/web.py
```diff
@@ -352,39 +352,43 @@ def signup_finish() -> ResponseReturnValue:  # noqa: PLR0911
     return redirect(url_for("root.index"))
 
 
+RMS_PROJECT_CREATION_FAILED_MESSAGE = (
```
zhmurov: We usually inline these

zhmurov (dismissed):

manytask/manytask/web.py
Ch0p1k3: Inlined in 56bfce9.

zhmurov (approved):

